PR #57, opened 2021-03-30
[BREAKING] fix(request_issue): make requested amount be the number of btc to be sent

sander2:
This changes request_issue such that the requested amount is equal to the amount of btc the user has to send. Note that this does mean that they will receive fewer PolkaBTC than they requested.

Changes are required in the UI. The testdata-gen tool needs to be updated as well, although it is not urgent - as far as I can see the current version has been broken for a long time, as it does not pay the fee


sander2:
Closed due to origin branch being wrong

PR #58, opened 2021-03-30
[BREAKING] fix(request_issue): make requested amount be the number of btc to be sent

sander2:
This changes request_issue such that the requested amount is equal to the amount of btc the user has to send. Note that this does mean that they will receive fewer PolkaBTC than they requested.

Changes are required in the UI. The testdata-gen tool needs to be updated as well, although it is not urgent - as far as I can see the current version has been broken for a long time, as it does not pay the fee

### DIFF
--- a/crates/issue/src/lib.rs
+++ b/crates/issue/src/lib.rs
@@ -174,7 +174,8 @@ decl_module! {
         /// # Arguments
         ///
         /// * `origin` - sender of the transaction
-        /// * `amount` - amount of PolkaBTC
+        /// * `amount` - amount of BTC the user wants to convert to PolkaBTC. Note that the amount of
+        /// PolkaBTC received will be less, because a fee is subtracted.
         /// * `vault` - address of the vault
         /// * `griefing_collateral` - amount of DOT
         #[weight = <T as Config>::WeightInfo::request_issue()]
@@ -246,7 +247,7 @@ impl<T: Config> Module<T> {
     /// Requests CBA issuance, returns unique tracking ID.
     fn _request_issue(
         requester: T::AccountId,
-        amount_polkabtc: PolkaBTC<T>,
+        amount_btc: PolkaBTC<T>,
         vault_id: T::AccountId,
         griefing_collateral: DOT<T>,
     ) -> Result<H256, DispatchError> {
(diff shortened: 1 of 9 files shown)


gregdhill (approved):

crates/issue/src/lib.rs
```diff
@@ -267,39 +269,37 @@ impl<T: Config> Module<T> {
...
 
-        let fee_polkabtc = ext::fee::get_issue_fee::<T>(amount_polkabtc)?;
-        let amount_btc = amount_polkabtc + fee_polkabtc;
-
-        let issue_id = ext::security::get_secure_id::<T>(&requester);
-
         ext::vault_registry::try_increase_to_be_issued_tokens::<T>(&vault_id, amount_btc)?;
 
+        let fee_polkabtc = ext::fee::get_issue_fee::<T>(amount_btc)?;
+        let user_polkabtc = amount_btc
```
nud3l, line 275: maybe rename to `amount_issued`

crates/issue/src/lib.rs
```diff
@@ -246,7 +247,7 @@ impl<T: Config> Module<T> {
     /// Requests CBA issuance, returns unique tracking ID.
     fn _request_issue(
         requester: T::AccountId,
-        amount_polkabtc: PolkaBTC<T>,
+        amount_btc: PolkaBTC<T>,
```
nud3l, line 250: maybe rename to `amount_requested`. Then we can slowly get rid of the btc and polkabtc references in the var names.

crates/issue/src/lib.rs
sander2: That's ambiguous, though - technically the vault issues both the fee and the amount sent to the user

nud3l: `amount_user`?

crates/issue/src/lib.rs
```diff
@@ -267,39 +269,39 @@ impl<T: Config> Module<T> {
...
         ext::collateral::lock_collateral::<T>(&requester, griefing_collateral)?;
 
-        let fee_polkabtc = ext::fee::get_issue_fee::<T>(amount_polkabtc)?;
-        let amount_btc = amount_polkabtc + fee_polkabtc;
+        ext::vault_registry::try_increase_to_be_issued_tokens::<T>(&vault_id, amount_requested)?;
 
-        let issue_id = ext::security::get_secure_id::<T>(&requester);
-
-        ext::vault_registry::try_increase_to_be_issued_tokens::<T>(&vault_id, amount_btc)?;
+        let fee_polkabtc = ext::fee::get_issue_fee::<T>(amount_requested)?;
```
nud3l, line 274: just `fee`?

nud3l (approved):

PR #59, opened 2021-03-30
 fix(request_redeem): remove redundant&wrong check, add tests for capacity limit

sander2:


crates/redeem/src/lib.rs
```diff
@@ -296,13 +296,8 @@ impl<T: Config> Module<T> {
             .checked_sub(&fee_polka_btc)
             .ok_or(Error::<T>::ArithmeticUnderflow)?;
 
-        let vault = ext::vault_registry::get_active_vault_from_id::<T>(&vault_id)?;
         let height = <frame_system::Pallet<T>>::block_number();
         ext::vault_registry::ensure_not_banned::<T>(&vault_id, height)?;
-        ensure!(
-            redeem_amount_polka_btc <= vault.issued_tokens,
```
sander2, line 303: This check was wrong, since it didn't account for to_be_redeemed tokens. This is checked anyway in the vault registry, so this one can safely be removed. I added an integration test ensuring that these limits are respected.

gregdhill (approved):

PR #60, opened 2021-03-31
fix: allow anyone to report vault theft

sander2:


crates/staked-relayers/src/lib.rs
```diff
@@ -502,8 +501,10 @@ decl_module! {
                 reports.insert(vault_id.clone());
             });
 
-            // reward relayer for this report by increasing its sla
-            ext::sla::event_update_relayer_sla::<T>(&signer, ext::sla::RelayerEvent::CorrectTheftReport)?;
+            // if the report is made by a relayer, reward it by increasing its sla
+            if Self::relayer_is_registered(&signer) {
```
sander2, line 505: Note: instead of only updating the SLA if the relayer is active, we now do it if the relayer is registered (i.e. active or inactive)

gregdhill (approved):

PR #86, opened 2021-04-20
Fix/off chain worker

sander2:
Tested with client integration test.

In the future we could add additional logic to ensure that at any time only a single off-chain worker is running, but I think for now this is fine


crates/btc-relay/src/lib.rs
```diff
@@ -317,6 +317,13 @@ decl_module! {
             Self::clear_block_error(block_hash, error)
         }
 
+        #[weight = 0]
+        #[transactional]
+        fn set_parachain_confirmations(origin, period: T::BlockNumber) {
```
sander2, line 322: This can be removed when Greg the `system.set_storage(key, value)` to the clients

gregdhill (approved):

PR #98, opened 2021-05-05
[DRAFT] adds RPC call to list all vaults with redeemable tokens

theodorebugnet:
Analogously to the existing get_vaults_with_issuable_tokens call.

Opened draft for feedback, will add tests shortly.


crates/vault-registry/src/lib.rs
```diff
@@ -1335,6 +1335,29 @@ impl<T: Config> Module<T> {
...
+                if !redeemable_tokens.is_zero() {
+                    Some((account_id, redeemable_tokens))
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<(_, _)>>();
+
+        if vaults_with_redeemable_tokens.is_empty() {
+            Err(Error::<T>::NoVaultWithRedeemableTokens.into())
```
nud3l, line 1354: A fix that we should probably do is to return an empty array instead of an error when there are 0 vaults. Because this is not an error state per see, it's just that we have 0 vaults right now. IMO we should also correct that behavior for issuable tokens.

parachain/runtime/src/lib.rs
```diff
@@ -873,6 +873,11 @@ impl_runtime_apis! {
             Ok(result.iter().map(|v| (v.0.clone(), BalanceWrapper{amount:v.1})).collect())
         }
 
+        fn get_vaults_with_redeemable_tokens() -> Result<Vec<(AccountId, BalanceWrapper<Balance>)>, DispatchError> {
+            let result = VaultRegistry::get_vaults_with_redeemable_tokens()?;
+            Ok(result.iter().map(|v| (v.0.clone(), BalanceWrapper{amount:v.1})).collect())
```
sander2, line 878: use `into_iter()` instead of `iter()`, that way you get ownership over the items you iterate over, so you won't have to do the clone in `v.0.clone()`.

Not that that this matters a huge deal, but since there are some other comments anyway I figured I'd include this :)

crates/vault-registry/src/tests.rs
```diff
@@ -1353,13 +1353,121 @@ mod get_vaults_with_issuable_tokens_tests {
...
+    }
+}
+
+mod get_vaults_with_redeemable_tokens_test {
+    use super::*;
+
+    #[test]
+    fn get_vaults_with_redeemable_tokens_returns_empty() {
+        run_test(|| {
+            create_sample_vault();
```
sander2, line 1367: Why is this here?

crates/vault-registry/src/tests.rs
```diff
@@ -1353,13 +1353,121 @@ mod get_vaults_with_issuable_tokens_tests {
...
+    fn get_vaults_with_redeemable_tokens_succeeds() {
+        run_test(|| {
+            let id1 = 3;
+            create_vault(id1);
+            let issued_tokens1: u128 = 10;
+            VaultRegistry::try_increase_to_be_issued_tokens(&id1, issued_tokens1).unwrap();
+            assert_ok!(VaultRegistry::issue_tokens(&id1, issued_tokens1));
+            let vault1 = VaultRegistry::get_active_rich_vault_from_id(&id1).unwrap();
+            assert_eq!(vault1.data.issued_tokens, issued_tokens1);
+            assert_eq!(vault1.data.to_be_redeemed_tokens, 0);
```
sander2, line 1383: Great to see these tests! They're a bit verbose though, maybe it would be nice to extract the 8 lines above into a reusable function inside of the get_vaults_with_redeemable_tokens_test module?

crates/vault-registry/src/lib.rs
```diff
@@ -1327,12 +1327,27 @@ impl<T: Config> Module<T> {
             })
             .collect::<Vec<(_, _)>>();
 
-        if vaults_with_issuable_tokens.is_empty() {
-            Err(Error::<T>::NoVaultWithIssuableTokens.into())
-        } else {
-            vaults_with_issuable_tokens.sort_by(|a, b| b.1.cmp(&a.1));
-            Ok(vaults_with_issuable_tokens)
-        }
+        vaults_with_issuable_tokens.sort_by(|a, b| b.1.cmp(&a.1));
```
sander2, line 1330: This change makes sense, but can you check that this won't break the UI/lib?

crates/vault-registry/src/tests.rs
theodorebugnet: To have a vault with 0 redeemable tokens, and make sure it doesn't get returned.

crates/vault-registry/src/lib.rs
theodorebugnet: I'll double-check, but there should already be handling. If there isn't I think it still makes more sense to fix the UI/lib to handle the empty array case than to throw an error here.

crates/vault-registry/src/tests.rs
sander2: Makes sense, can you add that as a comment?

nud3l (approved):

PR #241, opened 2021-08-26
fix: zero-cost base extrinsic weight, to be able to do runtime upgrade without fees

sander2:
This change is required to be able to do runtime upgrades as sudo without fees. As I understand it, the cost of a transaction is `TransactionByteFee * number_of_bytes_in_extrinsic + weights.base_extrinsic + weight_of_function`. While `sudo_unchecked_weight` can set the last part to zero, we need the other parts to be zero as well. Eventually we will want to revert this change for better spam protection.

One thing to note is that subxt hangs when trying to execute a no-cost function, so we have to use polkadot.js.

Finally, I had to compile with `--release` in order to get this to work, otherwise the block production took too long. We have to make sure that the kusama node is fast enough to execute in the allowed block time.

### DIFF
--- a/parachain/runtime/src/lib.rs
+++ b/parachain/runtime/src/lib.rs
@@ -137,7 +137,7 @@ parameter_types! {
     pub RuntimeBlockWeights: BlockWeights = BlockWeights::builder()
         .base_block(BlockExecutionWeight::get())
         .for_class(DispatchClass::all(), |weights| {
-            weights.base_extrinsic = ExtrinsicBaseWeight::get();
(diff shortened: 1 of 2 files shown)


sander2:
> Would it be sensible to add a test for this?

I am not sure how to test this. I wrote an integration test in the client, but as mentioned, subxt hangs so it is not really an option. In the parachain integration tests, I think functions don't incur a transaction fee at all

gregdhill (approved):
Approved, but we need to revert this ASAP!

PR #260, opened 2021-09-03
chore!: remove LiquidationVaultAccountId since it is outdated

sander2:
**Breaking changes**
- Storage
  - Removed `vault_registry::LiquidationVaultAccountId` - it is outdated (does not account for multicollateral). If necessary, lib can query `LiquidationVault`, which now also contains the `AccountId`


sander2:
Actually, it is actively used, and furthermore, it is not outdated at all - all liquidation vaults share the same account_id, but each uses a different currency within that account.

PR #305, opened 2021-10-06
test: Double-spending replace and refund

savudani8:


standalone/runtime/tests/mock/mod.rs
```diff
@@ -1310,3 +1312,75 @@ pub const fn wrapped(amount: u128) -> Amount<Runtime> {
 pub const fn griefing(amount: u128) -> Amount<Runtime> {
     Amount::new(amount, DOT)
 }
+
+pub fn request_replace(
```
sander2, line 1316: Can you place these functions in `mock/replace_testing_utils.rs`, akin to the redeem & issue helpers?

PR #417, opened 2021-12-10
tests: add basic annuity test for empty supply and add todo for annui…

nud3l:
…ty governance integration test


gregdhill (commented):
I think it might be worth doing the TODOs in this PR or splitting them out into a separate backlog ticket.

gregdhill (approved):

PR #456, opened 2021-12-18
testing branch

sander2:


parachain/src/chain_spec.rs
```diff
@@ -94,6 +94,7 @@ fn get_properties() -> Map<String, Value> {
     });
     properties.insert("tokenSymbol".into(), token_symbol.into());
     properties.insert("tokenDecimals".into(), token_decimals.into());
+	properties.insert("ss58Format".into(), 2092u16.into());
```
sander2, line 97: Apparently this needed, according to https://github.com/polkadot-js/api/issues/3864#issuecomment-896259716. Acala is also using this

PR #485, opened 2022-02-01
Chore: blacklist all xcm funcs

sander2:
Previously the xcm management functions were still allowed by the call filter, but because these are root-only functions, and sudo/governance ignores the filter, the whitelisting of these management functions was superfluous.


gregdhill (changes requested):
Clippy warning `unused variable: 'xcm_method'`

Do you actually want to revert the commit from [this PR](https://github.com/interlay/interbtc/pull/479)?

sander2:
> Clippy warning `unused variable: 'xcm_method'`
Fixed.

> Do you actually want to revert the commit from [this PR](https://github.com/interlay/interbtc/pull/479)?
No I don't - before that PR all xcm calls were allowed. After that PR, only some were. And now, none will be.

gregdhill (approved):

PR #497, opened 2022-02-04
fix: allow unpaid xcm messages in the testnet runtime

sander2:
Without this, we can't send tokens to the sibling chain in the rococo-local setup.


gregdhill (commented):
Conflict, please also see clippy warnings.

gregdhill (approved):

gregdhill (approved):

PR #506, opened 2022-02-12
making it work on m1 mac

r-medina:
this is almost certainly not what we want to do, but this draft PR is an example of how I finally got it to build on a Mac with

```sh
CC="zig cc" cargo check -p interlay-runtime-parachain
```

https://github.com/interlay/interbtc/issues/504


r-medina:
https://github.com/interlay/interbtc/issues/504

nud3l:
Hi! Thanks for opening this PR. Did you manage to resolve the issue in another way?

PR #589, opened 2022-05-04

sander2:
2 significant changes: 

- Only award a premium for the amount of tokens required to bring the vault back to the secure threshold. Otherwise the vault could end up paying the premium for _all_ of its tokens.
- Rather than only considering `issued_tokens` to determine whether or not a vault is below the premium redeem threshold, now use the "to_be_backed" tokens, i.e. the tokens if all open issues and redeems execute successfully (= issued + to_be_issued - to_be_redeemed). This is required because otherwise additional redeems would again be able to de a premium redeem. 

Since this significantly modifies audited code, I'd like to have 2 people approve this PR before we merge

### DIFF
--- a/crates/redeem/src/ext.rs
+++ b/crates/redeem/src/ext.rs
@@ -79,6 +79,12 @@ pub(crate) mod vault_registry {
         <vault_registry::Pallet<T>>::get_vault_from_id(vault_id)
     }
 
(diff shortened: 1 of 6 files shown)


sander2:
I just realized that by doing the premium redeem, the collateral decreases. I didn't account for that in the premium cap calculation - should I? That's going to complicate the calculation

PR #634, opened 2022-06-08
fix: update escrow period to double lease period, increase required vINTR for democracy and bump spec version

nud3l:
# Updates

- escrow.maxPeriod from 4,838,400 to 9676800
- democracy.minimumDeposit from 10,000,000,000 planck to 2,500,000,000,000 (250 vINTR required). That would allow about 7500 accounts to make a proposal if they stake their INTR for 2 years.

### DIFF
--- a/parachain/runtime/interlay/src/lib.rs
+++ b/parachain/runtime/interlay/src/lib.rs
@@ -88,7 +88,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("interlay-parachain"),
     impl_name: create_runtime_str!("interlay-parachain"),
     authoring_version: 1,
-    spec_version: 3, // 1.14.0
+    spec_version: 3, // 1.14.1
     impl_version: 1,
     transaction_version: 2, // added preimage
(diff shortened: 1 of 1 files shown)


parachain/runtime/interlay/src/lib.rs
```diff
@@ -88,7 +88,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("interlay-parachain"),
     impl_name: create_runtime_str!("interlay-parachain"),
     authoring_version: 1,
-    spec_version: 3, // 1.14.0
+    spec_version: 4, // 1.14.1
```
sander2, line 91: We could keep at 3 - afaik greg hasn't done the runtime upgrade yet, and changing to 4 would require another bump in the clients. Although we don't use the clients yet.. Approving this in case you don't want to make the change.

sander2 (dismissed):

parachain/runtime/interlay/src/lib.rs
nud3l: Agree, won't bump

sander2 (approved):

PR #647, opened 2022-06-16
feat(vault): liquidation recovery

savudani8:
Implements https://github.com/interlay/interbtc/issues/637 without allowing recovery with outstanding wrapped tokens, to avoid [this](https://github.com/interlay/interbtc/issues/637#issuecomment-1155453172) edge case.

### DIFF
--- a/crates/vault-registry/src/lib.rs
+++ b/crates/vault-registry/src/lib.rs
@@ -417,6 +417,26 @@ pub mod pallet {
             Self::_set_liquidation_collateral_threshold(currency_pair, threshold);
             Ok(())
         }
+
+        /// Recover vault ID from a liquidated status.
+        ///
+        /// # Arguments
+        /// * `currency_pair` - the currency pair to change
+        #[pallet::weight(<T as Config>::WeightInfo::recover_vault_id())]
+        #[transactional]
+        pub fn recover_vault_id(origin: OriginFor<T>, currency_pair: DefaultVaultCurrencyPair<T>) -> DispatchResult {
+            let account_id = ensure_signed(origin)?;
+            let vault_id = VaultId::new(account_id, currency_pair.collateral, currency_pair.wrapped);
+            ensure!(Self::is_vault_liquidated(&vault_id)?, Error::<T>::VaultNotRecoverable);
+
+            let mut vault = Self::get_rich_vault_from_id(&vault_id.clone())?;
+            ensure!(vault.to_be_redeemed_tokens().is_zero(), Error::<T>::VaultNotRecoverable);
+
+            // Vault accepts new issues by default
+            vault.set_accept_new_issues(true)?;
+
+            Ok(())
+        }
     }
 
     #[pallet::event]
@@ -560,6 +580,8 @@ pub mod pallet {
         VaultCommittedTheft,
         /// Vault is no longer usable as it was liquidated due to undercollateralization.
         VaultLiquidated,
+        /// Vault must be either liquidated or flagged for theft.
+        VaultNotRecoverable,
         /// No bitcoin public key is registered for the vault.
         NoBitcoinPublicKey,
         /// A bitcoin public key was already registered for this account.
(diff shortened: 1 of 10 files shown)


crates/vault-registry/src/types.rs
```diff
@@ -699,6 +699,14 @@ impl<T: Config> RichVault<T> {
         }
     }
 
+    pub fn has_no_outstanding_wrapped_tokens(&self) -> Result<bool, DispatchError> {
```
sander2, line 702: Imo having negations in function names can get confusing (especially if you ever need the negation anywhere) - I would flip the condition and go with `has_outstanding_wrapped_tokens`

crates/vault-registry/src/types.rs
```diff
@@ -699,6 +699,14 @@ impl<T: Config> RichVault<T> {
...
+        let total_wrapped_tokens = self
+            .backed_tokens()?
+            .checked_add(&self.to_be_replaced_tokens())?
```
sander2, line 705: I'm not sure this should be included here - to_be_replaced_tokens is not like issued/to_be_issued/to_be_redeemed, in that they don't represent an obligation - it's simply bookkeeping to keep track how many tokens the vault asked to be replaced for. 

Actually, now that I think about it, what we should probably do is to withdraw the replace request if there is one in the liquidation function. This will free up a little bit of collateral for the vault as well

crates/vault-registry/src/lib.rs
```diff
@@ -417,6 +417,32 @@ pub mod pallet {
...
+            ensure!(Self::is_vault_liquidated(&vault_id)?, Error::<T>::VaultNotRecoverable);
+
+            let mut vault = Self::get_rich_vault_from_id(&vault_id.clone())?;
+            ensure!(
+                vault.has_no_outstanding_wrapped_tokens()?,
+                Error::<T>::VaultNotRecoverable
+            );
+            vault.set_accept_new_issues(accept_new_issues)?;
+
+            Ok(())
```
sander2, line 444: Should we clear the `banned_until` field?

sander2 (commented):
Can you add a test where you add collateral after recovery and actually issue some tokens?

crates/vault-registry/src/types.rs
savudani8: It's not as straightforward as calling `withdraw_replace()` in the `replace` pallet, since `replace` depends on the `vault-registry` rather than the other way around. Should I move the function [body](https://github.com/interlay/interbtc/blob/825030f2caaf741c811568b24d42d1a2547810f7/crates/replace/src/lib.rs#L361) inside the vault registry?

sander2: Yea I think we can move [these lines](https://github.com/interlay/interbtc/blob/825030f2caaf741c811568b24d42d1a2547810f7/crates/replace/src/lib.rs#L364-L372) to a `withdraw_replace_request` function in the vault registry

standalone/runtime/tests/test_vault_registry.rs
```diff
@@ -343,3 +363,144 @@ fn integration_test_vault_registry_register_respects_fund_limit() {
...
+                vault.issued = vault_id.wrapped(0);
+                vault.to_be_issued = vault_id.wrapped(0);
+                vault.to_be_redeemed = vault_id.wrapped(0);
+                vault.to_be_replaced = vault_id.wrapped(0);
+                vault.backing_collateral = Amount::new(0, currency_id);
+                vault.liquidated_collateral = Amount::new(0, currency_id);
+                *vault
+                    .free_balance
+                    .get_mut(&vault_data.replace_collateral.currency())
+                    .unwrap() += vault_data.replace_collateral;
```
sander2, line 501: Why did the vault not receive the collateral that was above [the liquidation threshold at the time of liquidation]?

sander2 (commented):
Could you also add a test where you try do the liquidation recovery _after you execute all open redeems_? Would it also be good to test the same after _cancelling_ the redeem? What do you think?

standalone/runtime/tests/test_vault_registry.rs
savudani8: That's because in this setup the vault was undercollateralized. I realised this test wasn't testing changes caused by the actual recovery extrinsic, so I modified it slightly.

savudani8:
> Could you also add a test where you try do the liquidation recovery after you execute all open redeems? Would it also be good to test the same after cancelling the redeem? What do you think?

Good to have these tests in case the underlying `to_be_redeemed` tokens check ever becomes insufficient, I added both.

sander2 (approved):
LGTM, but 2 notes. First, it'd be good to emit an event so the vault client can listen to it, and start functioning after receiving it. I would be ok with leaving that for a future PR if you prefer to merge this soon (as I realize it's been open for a while now). 
Second, I think it'd be good to have @gregdhill give this a once-over as well before we merge

gregdhill (approved):
LGTM!

PR #650, opened 2022-06-20
feat: pending vault theft status

savudani8:
Adds `PendingTheft` to the `VaultStatus` enum. Part of the implementation for https://github.com/interlay/interbtc/issues/636.


sander2:
I don't think we should merge this pr in isolation - I'd rather have 1 pr that includes the whole recover-from-theft feature

PR #656, opened 2022-06-30
Vault Theft Recovery

savudani8:
Implements https://github.com/interlay/interbtc/issues/636
Depends on https://github.com/interlay/interbtc/pull/647

- New vault status: `PendingTheft(amount, recovery_deadline_block)`. It is equivalent to being banned indefinitely.
- The rewarding of theft reporting is now separated from vault liquidation. `relay::report_vault_theft(...)` and  `relay::report_vault_double_payment(...)` call `vault_registry::report_vault_theft(...)` which distributes the theft fee to the reporter and either moves a vault from status `Active(...)` to `PendingTheft(...)`, or from `PendingTheft(amount, deadline)` to `PendingTheft(amount + new_theft, deadline)`.
- `amount` in the `PendingTheft(...)` status is computed as the total output of the reported transaction. This overestimates the stolen amount, but is much simpler than having input transactions reported and adding up only amounts from the vault's wallet.
- New extrinsic: `recover_vault_theft(...)`. It reduces the amount in `PendingTheft(amount, deadline)` by the sum of outputs to the vault's wallet. No input must originate from the vault's wallet. If `amount` reaches zero, the vault becomes `Active(true)`.
- New extrinsic: `pending_liquidation_elapsed(...)`. Called by the offchain worker, moves a vault from `PendingTheft(...)` status to `CommittedTheft` and liquidates it.
- The `VaultTheft` event is updated to include the stolen amount. `VaultDoublePayment` is left as-is, since it steals the same amount as in the request.
- New event: `VaultTheftRecovery`

### DIFF
--- a/crates/issue/src/lib.rs
+++ b/crates/issue/src/lib.rs
@@ -276,6 +276,9 @@ impl<T: Config> Pallet<T> {
             Error::<T>::WaitingForRelayerInitialization
         );
 
+        // Check that the vault is currently not banned or pending theft
+        ext::vault_registry::ensure_not_banned::<T>(&vault_id)?;
+
         let vault = ext::vault_registry::get_active_vault_from_id::<T>(&vault_id)?;
 
         // ensure that the vault is accepting new issues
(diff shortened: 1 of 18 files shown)


savudani8:
Closing for now.

PR #657, opened 2022-07-01
fix: ensure that report double payment transactions are unique

nud3l:


sander2 (approved):
Happy to have this fix in asap, but in a separate PR we should also address the transaction parsing

PR #678, opened 2022-07-26
fix: avoid overwriting changes in liquidate()

sander2:
I stumbled upon this indirectly when I added some integrity checks. The actual fix is in `vault-registry/src/types.rs` - the `withdraw_replace_request` function would change the vault struct, but then the change got overwritten later on


gregdhill (commented):
Good catch!

Issue #743, opened 2022-11-02
Enforce collateral toggle on deposits


savudani8:
**Is your feature request related to a problem? Please describe.**
The flow introduced by using orml-tokens in the lending protocol complicates the user flow, and may be unfamiliar to users. 

**Describe the solution you'd like**
Enforce the "collateral toggle" model on deposits:
- [ ] Require locking the entire LendToken balance at once as collateral
    - If the origin account already has locked LendTokens, the toggle-on tx should fail. If tokens are locked elsewhere (for instance, as Vault collateral), enabling the collateral toggle wouldn't be an accurate representation of the underlying state (because some LendTokens will be locked in the lending protocol while some will be locked as Vault collateral).
- [ ] New deposits of underlying or incoming LendToken transfers are automatically locked as collateral if the user already has a deposit into the lending pool (so the collateral toggle is on).
- [x] Unlocking collateral from the lending pool can only happen all-at-once.
- [ ] Enable users to redeem part of the locked LendTokens in a single tx, by doing the unlocking step under-the-hood, if their loans would still be backed.
    - ~I'm not convinced this needs to be implemented on the parachain. The calls could be batched by the [interbtc-api](https://github.com/interlay/interbtc-api/).~
    - This is standard lending UX
    - The tokens have to be redeemed directly rather than only unlocked, because unlocking only a part of the LendTokens would bring the user to an inconsistent state: collateral is neither entirely toggled on, nor toggled off.

PR #754, opened 2022-11-07
Collateral Toggle

savudani8:
- Auto-deposits incoming lend tokens from minting and transfers, if the user already enabled the toggle
    - Implemented using `orml-tokens` post-hooks for OnDeposit and OnTransfer
    - The `deposit_collateral(...)` extrinsic can now be removed 
- `redeem(...)` implicitly calls  `withdraw(...)` if the user locked funds (the collateral toggle is enabled). 
    - Thanks to this, the `withdraw_collateral(...)` extrinsic is now removed.
- Forbids enabling the lend tokens as collateral unless the entire balance of the user is `free`

Closes https://github.com/interlay/interbtc/issues/743

TODO:
- [x] upgrade to ORML to https://github.com/open-web3-stack/open-runtime-module-library/commit/5803a01a44ade05fe5f00dd73b619c89fbf4e701

### DIFF
--- a/crates/loans/src/lib.rs
+++ b/crates/loans/src/lib.rs
@@ -40,7 +40,7 @@ use frame_support::{
 };
 use frame_system::pallet_prelude::*;
 use num_traits::cast::ToPrimitive;
-use orml_traits::MultiCurrency;
+use orml_traits::{MultiCurrency, MultiReservableCurrency};
 pub use pallet::*;
 use primitives::{Balance, CurrencyId, Liquidity, Price, Rate, Ratio, Shortfall, Timestamp};
 use sp_runtime::{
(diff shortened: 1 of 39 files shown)


crates/loans/src/lib.rs
```diff
@@ -950,22 +997,13 @@ pub mod pallet {
...
-        pub fn deposit_collateral(
-            origin: OriginFor<T>,
-            asset_id: AssetIdOf<T>,
-            #[pallet::compact] amount: BalanceOf<T>,
-        ) -> DispatchResultWithPostInfo {
-            let who = ensure_signed(origin)?;
-            ensure!(!amount.is_zero(), Error::<T>::InvalidAmount);
-            Self::do_deposit_collateral(&who, asset_id, amount)?;
+            let free_lend_tokens = Self::free_lend_tokens(asset_id, &who)?;
+            ensure!(!free_lend_tokens.is_zero(), Error::<T>::InvalidAmount);
```
sander2, line 1001: We should probably return a more descriptive error here - it'll be confusing for users since they did not specify any amount. Also, this error will also be thrown if they already called `deposit_all_collateral` previously, right?

crates/loans/src/lib.rs
```diff
@@ -950,22 +997,13 @@ pub mod pallet {
...
+            let reserved_lend_tokens = Self::reserved_lend_tokens(asset_id, &who)?;
+            // This check could fail if `withdraw_all_collateral()` leaves leftover lend_tokens locked.
+            // However the current implementation is guaranteed to withdraw everything.
+            ensure!(reserved_lend_tokens.is_zero(), Error::<T>::TokensAlreadyLocked);
+            Self::do_deposit_collateral(&who, free_lend_tokens.currency(), free_lend_tokens.amount())?;
```
sander2, line 1006: Let's refactor `do_deposit_collateral` to `do_deposit_all_collateral`, and move the logic above this to inside that function? Since we can't call `do_deposit_collateral` with arbitrary amounts anymore

savudani8: We would end up with a `do_deposit_all_collateral` and a `do_withdraw_collateral`, because the latter is still needed on partial redeems. I'm not sure how much `do_deposit_all_collateral` adds for sacrificing the current uniform interface (where `do_deposit_collateral()` and `do_withdraw_collateral()` are clearly opposites).

crates/loans/src/lib.rs
savudani8: It's not the best worded error but it was already in use everywhere else in the pallet and its doc comment is `Amount cannot be zero`. Maybe I should just rename the error to `AmountIsZero`?

sander2: But my point is that `Amount cannot be zero` is confusing since the user is not supplying any explicit amount as argument.

Ideally we'd have something like `AlreadyConfiguredAsCollateral` when they already have the toggle enabled, or `BalanceZero` (with docstring like "Operation unsupported when balance is zero.") when they just try to call this without actually having the token

savudani8: This is something the UI and lib are on the way of preventing as part of input validation, but I did rephrase the error messages.

gregdhill (approved):

PR #767, opened 2022-11-14
feat: xcm taxing

sander2:
Proof of concept of taxing xcm transfers. Obviously needs a lot of cleanup before we can consider merging this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1936,7 +1936,6 @@ dependencies = [
  "mocktopus",
  "orml-tokens",
  "orml-traits",
- "pallet-traits",
  "pallet-transaction-payment",
  "parity-scale-codec",
  "scale-info",
@@ -1946,6 +1945,7 @@ dependencies = [
  "sp-io",
  "sp-runtime",
  "sp-std",
+ "traits",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


nud3l:
>     * [ ]  make taxing rate configurable. The easiest would be to use the same rate for all tokens, inbound or outbound, regardless of whether those tokens are native to our chain. But if we want settings per direction and/or token we can implement that as well.
>       
>       * [ ]  decide on requirements
>       * [ ]  implement it

It's a bit of a tricky question, but I think we only want to tax kBTC/iBTC. Even when iBTC/kBTC are provided to lending or LPd in the DEX, they are but to use on the Interlay/Kintsugi networks so I don't see a reason to impose extra tax on them.

>     * [ ]  Transfer taxes to the reward pools rather than to the treasury account

Transfer to treasury is good. Vaults mostly care about the block rewards.

>     * [ ]  Decide what to do with non-native currencies - current reward/staking logic only works for our native tokens

Native tokens only is sufficient.

>     * [ ]  Possibly exempt governance/treasury transfers from taxes?

For simplicity, I think we should tax everyone the same.

>     * [ ]  Possibly write a script to inspect all past xcm messages to see if any would have failed with the new logic

Might be nice to do that. Most effort will be to get all partners to update the fee estimation in their UIs/wallets.

Issue #744, opened 2022-11-02
Replace usage of `get_price()` with `amount.convert()`


savudani8:
To reduce boilerplate code in the oracle integration of InterLend, replace all usage of `get_price()` with `amount.convert()`, which is the approach used in all our other parachain pallets.

EDIT: this also fixes a critical oracle integration bug where collateral is worth more if its currency has more decimals

Issue #751, opened 2022-11-04
Use fail-safe arithmetic in lending protocol


daniel-savu:
The default implementation doesn't use checked or saturated math.

PR #785, opened 2022-11-21
fix(loans): oracle integration

savudani8:
Replaces the incorrect `get_price` integration with the conversion approach from the `currency` crate. The `PriceFeeder` associated type is replaced by `ReferenceAssetId`, which is the currency that prices are expressed in by the oracle. In our runtimes, that currency is `GetWrappedCurrencyId`. 

In some tests, `KINT` and `KBTC` are replaced with `INTR` and `IBTC`, because runtime currencies have been harmonized for these pairs: governance_currency / loan_reward_currency, and wrapped_currency / reference_currency.

Closes https://github.com/interlay/interbtc/issues/744

### DIFF
--- a/crates/loans/src/lend_token.rs
+++ b/crates/loans/src/lend_token.rs
@@ -135,16 +135,12 @@ impl<T: Config> Pallet<T> {
 
         // Formula
         // reducible_underlying_amount = liquidity / collateral_factor / price
-        let price = Self::get_price(underlying_id)?;
-
         let reducible_supply_value = liquidity
             .checked_div(&market.collateral_factor.into())
             .ok_or(ArithmeticError::Overflow)?;
-
-        let reducible_underlying_amount = reducible_supply_value
-            .checked_div(&price)
-            .ok_or(ArithmeticError::Underflow)?
-            .into_inner();
+        let reducible_supply_amount =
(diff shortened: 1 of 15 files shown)


crates/loans/src/lend_token.rs
```diff
@@ -135,16 +135,12 @@ impl<T: Config> Pallet<T> {
...
         let reducible_supply_value = liquidity
             .checked_div(&market.collateral_factor.into())
             .ok_or(ArithmeticError::Overflow)?;
-
-        let reducible_underlying_amount = reducible_supply_value
-            .checked_div(&price)
-            .ok_or(ArithmeticError::Underflow)?
-            .into_inner();
+        let reducible_supply_amount =
+            Amount::<T>::from_unsigned_fixed_point(reducible_supply_value, T::ReferenceAssetId::get())?;
```
sander2, line 142: Would it be possible to have `get_account_liquidity`  return an `Amount`?

savudani8: I just tried it now but it has a lot of ramifications. Switching to your suggestion fixes https://github.com/interlay/interbtc/issues/751 and I'll do this after the testnet launch.

sander2 (approved):

Issue #791, opened 2022-11-29
Add initialization for min_exchange_rate and max_exchange_rate


savudani8:
While these fields are set in the genesis config, if lending is added via runtime upgrade these fields will default to zero. This would make lending unusable, with an underflow / overflow error.

Fix: Add initialization logic that runs on the runtime upgrade and sets those fields.

PR #799, opened 2022-12-06
fix(loans): Set min and max exchange rates on runtime upgrade

savudani8:
Adds a migration to set the `min` and `max` exchange rates, if these are zero.

Closes https://github.com/interlay/interbtc/issues/791

### DIFF
--- a/.github/workflows/cargo-test.yml
+++ b/.github/workflows/cargo-test.yml
@@ -31,6 +31,11 @@ jobs:
         with:
           command: check
           args: --release --workspace
+      - name: try-runtime build
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --release --workspace --features try-runtime
       - name: test
         uses: actions-rs/cargo@v1
         with:
--- a/.github/workflows/cargo-test.yml
+++ b/.github/workflows/cargo-test.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           command: check
           args: --release --workspace
-      - name: try-runtime build
+      - name: try-runtime check
         uses: actions-rs/cargo@v1
         with:
           command: check
--- a/crates/loans/src/lib.rs
+++ b/crates/loans/src/lib.rs
@@ -77,6 +77,9 @@ pub mod weights;
 pub const REWARD_ACCOUNT_PREFIX: &[u8; 13] = b"loans/farming";
 pub const INCENTIVE_ACCOUNT_PREFIX: &[u8; 15] = b"loans/incentive";
 
+pub const DEFAULT_MAX_EXCHANGE_RATE: u128 = 1_000_000_000_000_000_000; // 1
+pub const DEFAULT_MIN_EXCHANGE_RATE: u128 = 20_000_000_000_000_000; // 0.02
+
 type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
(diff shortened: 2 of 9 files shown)


.github/workflows/cargo-test.yml
```diff
@@ -31,6 +31,11 @@ jobs:
         with:
           command: check
           args: --release --workspace
+      - name: try-runtime build
```
sander2, line 34:
```suggestion
      - name: try-runtime check
```

crates/loans/src/lib.rs
```diff
@@ -343,6 +346,40 @@ pub mod pallet {
         IncentiveReservesReduced(T::AccountId, AssetIdOf<T>, BalanceOf<T>),
     }
 
+    #[pallet::hooks]
+    impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+        fn on_runtime_upgrade() -> frame_support::weights::Weight {
```
sander2, line 351: Please add a guard that checks (and updates) the `palletVersion` so this is only executed once. Note that `palletVersion` is part of Frame - we want to move away from the `StorageVersion` enum that we were using until now

sander2: Hmm yea there is advantage of using the storage version, namely that you can easily see if the migration runs or not. But fair enough about it already been guaranteed to only run once

crates/loans/src/lib.rs
```diff
@@ -343,6 +346,40 @@ pub mod pallet {
...
+            let max_exchange_rate = crate::MaxExchangeRate::<T>::get();
+            if max_exchange_rate.is_zero() {
+                crate::MaxExchangeRate::<T>::put(Rate::from_inner(DEFAULT_MAX_EXCHANGE_RATE));
+            }
+            let min_exchange_rate = crate::MinExchangeRate::<T>::get();
+            if min_exchange_rate.is_zero() {
+                crate::MinExchangeRate::<T>::put(Rate::from_inner(DEFAULT_MIN_EXCHANGE_RATE));
+            }
+            T::DbWeight::get().reads_writes(0, 2)
```
sander2, line 360: Nit: should be this, but it doesn't matter a ton
```suggestion
            T::DbWeight::get().reads_writes(2, 2)
```

crates/loans/src/lib.rs
savudani8: We can check whether the migration ran if the `MinExchangeRate` and `MaxExchangeRate` storage items are non-zero

parachain/runtime/testnet-kintsugi/src/lib.rs
```diff
@@ -1123,19 +1123,8 @@ pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<Address, RuntimeCall,
...
-    AllPalletsWithSystem,
-    (
-        // "Bound uses of call" <https://github.com/paritytech/substrate/pull/11649>
-        pallet_preimage::migration::v1::Migration<Runtime>,
-        pallet_scheduler::migration::v3::MigrateToV4<Runtime>,
-        pallet_multisig::migrations::v1::MigrateToV1<Runtime>,
-    ),
->;
+pub type Executive =
+    frame_executive::Executive<Runtime, Block, frame_system::ChainContext<Runtime>, Runtime, AllPalletsWithSystem>;
```
sander2, line 1127: we should also remove the kintsugi mainnet migrations but we can do that in another pr

sander2 (approved):

PR #805, opened 2022-12-12
feat: table two proposals on monday

### LINKED ISSUE #803: Table 2 referenda on Monday
Due to block time variation, it is hard to plan when proposals are tabled to a referendum. Moreover, since only a single proposal is tabled in a given period, there is a backlog of referenda.

1. Table referenda on Monday at 09.00 am UTC. Assuming 2 days of voting on kintsugi and 7 days of voting on interlay, this will mean that referenda are enacted on Wednesday noon-ish on kintsugi and Monday noon-ish on interlay. This is when most of the Interlay team is online to help with any potential issues.
2. Table 2 referenda each period. This will allow proposals to pass in parallel, e.g., one treasury and one regular governance proposal.

sander2:
Tables 2 proposals on week boundaries, i.e. monday morning at 9:00 UTC.

Implementation details: 
- stores the next boundary rather than recalculating it every block. This will be better performance-wise, and will make it easier for external tools to determine when tabling happens
- uses `chrono`
- after running the runtime upgrade, the `NextLaunchTimestamp` will be zero, meaning tabling will happen immediately
- Every time launching should happen, the next one is scheduled to be 7 days after the start of the current week. 
- We don't consider leap seconds since that kind of granularity is not required for us
- If block production stops for whatever reasons, and the week boundary is passed in the meantime, the next block will table the proposals regardless of the day
- Referendum duration remains a fixed number of blocks, so the enactment date still depends on the blocktime
- if launching happens on monday before 9 AM, the next launch will be next week 9AM, (rather than today 9AM). So in this case there will be slightly more than 7 days between tablings

Closes #803

### DIFF
--- a/crates/democracy/src/benchmarking.rs
+++ b/crates/democracy/src/benchmarking.rs
@@ -42,7 +42,7 @@ fn add_referendum<T: Config>(n: u32) -> Result<ReferendumIndex, &'static str> {
     let proposal_hash: T::Hash = T::Hashing::hash_of(&n);
     let vote_threshold = VoteThreshold::SimpleMajority;
 
-    Democracy::<T>::inject_referendum(T::LaunchPeriod::get(), proposal_hash, vote_threshold, 0u32.into());
+    Democracy::<T>::inject_referendum(T::VotingPeriod::get(), proposal_hash, vote_threshold, 0u32.into());
     let referendum_index: ReferendumIndex = ReferendumCount::<T>::get() - 1;
     T::Scheduler::schedule_named(
         (DEMOCRACY_ID, referendum_index).encode(),
@@ -198,7 +198,7 @@ benchmarks! {
         // Launch public
         assert!(add_proposal::<T>(r).is_ok(), "proposal not created");
 
-        let block_number = T::LaunchPeriod::get();
+        let block_number = T::VotingPeriod::get();
 
     }: { Democracy::<T>::on_initialize(block_number) }
     verify {
@@ -264,7 +264,7 @@ benchmarks! {
         assert_eq!(Democracy::<T>::referendum_count(), r, "referenda not created");
         assert_eq!(Democracy::<T>::lowest_unbaked(), 0, "invalid referenda init");
 
-        let block_number = T::LaunchPeriod::get();
+        let block_number = T::VotingPeriod::get();
 
     }: { Democracy::<T>::on_initialize(block_number) }
     verify {
--- a/crates/democracy/src/lib.rs
+++ b/crates/democracy/src/lib.rs
@@ -80,12 +80,16 @@
 #![recursion_limit = "256"]
 #![cfg_attr(not(feature = "std"), no_std)]
 
+use core::time::Duration;
+
+use chrono::{Days, NaiveTime};
 use codec::{Decode, DecodeLimit, Encode, Input, MaxEncodedLen};
 use frame_support::{
     ensure,
     traits::{
         schedule::{DispatchTime, Named as ScheduleNamed},
         BalanceStatus, Currency, Get, LockIdentifier, OnUnbalanced, ReservableCurrency, UnfilteredDispatchable,
+        UnixTime,
     },
     transactional,
     weights::Weight,
@@ -157,6 +161,8 @@ enum Releases {
 
 #[frame_support::pallet]
 pub mod pallet {
+    use core::num::TryFromIntError;
+
     use super::*;
     use frame_support::{
         dispatch::{DispatchClass, DispatchResultWithPostInfo, Pays},
@@ -190,10 +196,6 @@ pub mod pallet {
         /// where they are on the losing side of a vote.
         #[pallet::constant]
         type EnactmentPeriod: Get<Self::BlockNumber>;
-
-        /// How often (in blocks) new public referenda are launched.
-        #[pallet::constant]
-        type LaunchPeriod: Get<Self::BlockNumber>;
 
         /// How often (in blocks) to check for new votes.
         #[pallet::constant]
@@ -238,6 +240,9 @@ pub mod pallet {
         /// The maximum number of public proposals that can exist at any time.
         #[pallet::constant]
         type MaxProposals: Get<u32>;
+
+        /// Unix time
+        type UnixTime: UnixTime;
     }
 
     /// The number of (public) proposals that have been made so far.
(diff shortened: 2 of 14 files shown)


nud3l:
I think best 9:00 AM UTC as mentioned here: #803

crates/democracy/src/benchmarking.rs
```diff
@@ -42,7 +42,7 @@ fn add_referendum<T: Config>(n: u32) -> Result<ReferendumIndex, &'static str> {
     let proposal_hash: T::Hash = T::Hashing::hash_of(&n);
     let vote_threshold = VoteThreshold::SimpleMajority;
 
-    Democracy::<T>::inject_referendum(T::LaunchPeriod::get(), proposal_hash, vote_threshold, 0u32.into());
+    Democracy::<T>::inject_referendum(T::VotingPeriod::get(), proposal_hash, vote_threshold, 0u32.into());
```
nud3l, line 45: Is that actually the voting period? I though the launch period was basically the period where people could second a proposal before it is a referendum and afterward it becomes a referendum and people can vote on it.

sander2: The launch period was the interval at which proposals would get automatically tabled. The `VotingPeriod` is actually the right argument here, since it's the block number at which the referendum is supposed to end. But anyway that doesn't really matter here - it's only benchmarking code, and the actual value doesn't impact the weight.

sander2:
> I think best 9:00 AM UTC as mentioned here: #803

Done. This adds one new edge case though: we need to decide what to do if launching happens on monday before 9 AM. The behavior I implemented is that that next launch will be next week 9AM, (rather than today 9AM). So in this case there will be slightly more than 7 days between tablings

crates/democracy/src/lib.rs
```diff
@@ -1069,11 +1085,14 @@ impl<T: Config> Pallet<T> {
         let r = last.saturating_sub(next);
 
         // pick out another public referendum if it's time.
-        if (now % T::LaunchPeriod::get()).is_zero() {
+        let current_time = T::UnixTime::now();
```
gregdhill, line 1088: To confirm, this reads from `pallet-timestamp` right?

crates/democracy/src/lib.rs
```diff
@@ -1091,6 +1110,34 @@ impl<T: Config> Pallet<T> {
...
+
+        // time to launch - calculate the date of next launch.
+
+        // convert to format used by `chrono`
+        let secs: i64 = now.as_secs().try_into()?;
+        let now =
+            chrono::NaiveDateTime::from_timestamp_opt(secs, now.subsec_nanos()).ok_or(Error::<T>::TryIntoIntError)?;
+
+        // calculate next week boundary
+        let beginning_of_week = now.date().week(chrono::Weekday::Mon).first_day();
```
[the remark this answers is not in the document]
gregdhill: IMO the date and time should be parameterizable.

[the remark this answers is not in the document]
sander2: > I would prefer to keep LaunchPeriod and use that to store chrono::Weekday::Mon.

That doesn't make sense - it wouldn't be a launch period, the period would always be 7 days. It would only change the launch weekday.

> IMO the date and time should be parameterizable.

I disagree - why complicate things unnecessarily? Imo it's better to keep things as simple as possible

crates/democracy/src/lib.rs
sander2: yes

crates/democracy/src/lib.rs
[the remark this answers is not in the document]
gregdhill: My primary dislike is that it is effectively a magic number buried in the code, can we at least make it a constant?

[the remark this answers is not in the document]
nud3l: It would be good if we could read the time somehow via polkadot.js e.g., as a constant because otherwise subsquare or others might need to hardcode it cand can't get it programmatically easily.

Having it configurable would also be OK for me but not a strict requirement.

crates/democracy/src/lib.rs
```diff
@@ -1091,6 +1115,41 @@ impl<T: Config> Pallet<T> {
...
+
+        // convert to format used by `chrono`
+        let secs: i64 = now.as_secs().try_into().map_err(Error::<T>::from)?;
+        let now =
+            chrono::NaiveDateTime::from_timestamp_opt(secs, now.subsec_nanos()).ok_or(Error::<T>::TryIntoIntError)?;
+
+        // calculate next week boundary
+        let beginning_of_week = now.date().week(chrono::Weekday::Mon).first_day();
+        let next_week = beginning_of_week
+            .checked_add_days(Days::new(7))
```
gregdhill, line 1135: Would be good if the launch offset variable included this, as @nud3l pointed out we want to make it easy for downstream consumers to just so last timestamp + offset

sander2: That doesn't make any sense - the offset constant determines the _schedule_, i.e. at which point each week the proposals get launched. Including 7 days offset in that is nonsensical. That'd be like saying that you wake up every morning at 32:00 rather than at 8:00. If you just want to fetch the concrete timestamp at which the next proposal gets launched, you can read the `NextLaunchTimestamp` storage item - no offset reading required there

gregdhill: Not sure I follow your example but in my head what makes the most sense is having this recurring boundary like the launch period, but I'm not going to dispute this for the sake of brevity. Also fwiw `NextLaunchTimestamp` will be unset when we enact this but I guess that is fine since it will just launch whatever public proposal is next.

gregdhill (approved):

crates/democracy/src/lib.rs
gregdhill: So this is what you have implemented:
```
LaunchOffsetMillis = 9 * 60 * 60 * 1000 // 9 hours
MondayTimestamp = now() - num_days_from_monday()
NextLaunchTimestamp = MondayTimestamp + 7 + LaunchOffsetMillis
```

What I am proposing is something like this:
```
LaunchOffsetMillis = 60 * 60 * 168 * 1000 // one week
NextLaunchTimestamp = NextLaunchTimestamp + LaunchOffsetMillis
```

This assumes `NextLaunchTimestamp` is correctly set to `MondayTimestamp` as calculated in your implementation.

gregdhill: To calculate the `NextLaunchTimestamp` in `n` weeks using my approach it is simply `NextLaunchTimestamp + LaunchOffsetMillis * n`, whereas in your approach it would be `MondayTimestamp + 7 * n + LaunchOffsetMillis` - where `MondayTimestamp` has to be calculated on the fly.

PR #846, opened 2023-01-11
feat: true oracle medianizing

sander2:
Since we want to move toward having multiple oracle submitters, I think it's a good idea to have true medianizing rather than using the middle value. My reasoning is:
- the middle value was potentially unstable - given an input of even length, it would be very bad if we swap unpredictably between the two middle values
- with the true median, changes are a little bit smoother when the number of active oracles changes
- the number of authorized oracles is expected to be not very great, making the increased complexity of O(n*log(n)) (compared to O(n) for the middle value) not really important

### DIFF
--- a/crates/oracle/src/lib.rs
+++ b/crates/oracle/src/lib.rs
@@ -366,20 +366,20 @@ impl<T: Config> Pallet<T> {
     }
 
     fn median(mut raw_values: Vec<UnsignedFixedPoint<T>>) -> Option<UnsignedFixedPoint<T>> {
-        let mid_index = raw_values.len() / 2;
+        let mid_index = raw_values.len().checked_div(2)?;
         raw_values.sort_unstable();
         match raw_values.len() {
             0 => None,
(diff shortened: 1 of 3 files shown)


crates/oracle/src/lib.rs
```diff
@@ -353,16 +353,33 @@ impl<T: Config> Pallet<T> {
...
             if let OracleKey::ExchangeRate(currency_id) = key {
                 T::OnExchangeRateChange::on_exchange_rate_change(currency_id);
             }
 
-            Some(value.value)
+            Some(value)
+        }
+    }
+
+    fn median(mut raw_values: Vec<UnsignedFixedPoint<T>>) -> Option<UnsignedFixedPoint<T>> {
```
nud3l, line 368: Safe math in this function?

nud3l (commented):
Apart from the minor comment to use safe math throughout, LGTM!

nud3l (approved):

PR #850, opened 2023-01-12
feat(loans): remove `AccountEarned` storage

daniel-savu:
`AccountEarned` is only stored for helping the UI and the way it is computed is wrong. To fix it, we would need to call `accrue_interest()` on each Lend Token transfer, along with the computation that actually updates `AccountEarned`. I am removing this because it's inaccurate and the fix is too computationally expensive.

The formula is correct in Parallel's case because they don't allow users to transfer lend tokens.


sander2:
If we don't have plans to show this in the ui i'm happy to remove this

sander2 (approved):
approving under the assumption that this is not a requirement from the UI side

Issue #848, opened 2023-01-12
Integrate Lend Tokens with the capacity model


daniel-savu:
The `on_exchange_rate_change` function in the loans pallet must be called whenever interest is accrued so that the `Capacity` reward pool is updated. This is required for incentivising Vaults with Lend Token collateral.

PR #853, opened 2023-01-13
feat(loans): update capacity model on exchange rate updates

daniel-savu:
Closes https://github.com/interlay/interbtc/issues/848

Adds a lend_token vault to the existing nomination rewards test, and also adds an integration test to check that accrued interest in lend_tokens is results in a higher stake of the lend_token currency in the Capacity pool.

FYI: Vaults will have their capacity model reward for Lend Tokens updated only when someone interacts with that market, but that should be often enough in practice.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3811,6 +3811,7 @@ dependencies = [
  "sp-runtime",
  "sp-session",
  "sp-std",
+ "sp-timestamp",
  "sp-transaction-pool",
  "sp-version",
  "staking",
@@ -5094,6 +5095,7 @@ dependencies = [
  "sp-runtime",
  "sp-std",
  "traits",
+ "visibility",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


daniel-savu:
I still need to add an integration test where accrued `lend_token` interest increases a vault's rewards, so not ready to merge yet

sander2:
> FYI: Vaults will have their capacity model reward for Lend Tokens updated only when someone interacts with that market, but that should be often enough in practice.

Can you elaborate on this? What actions result in changes in the exchange rate that don't immediately update the capacity model?

daniel-savu:
> > FYI: Vaults will have their capacity model reward for Lend Tokens updated only when someone interacts with that market, but that should be often enough in practice.
> 
> Can you elaborate on this? What actions result in changes in the exchange rate that don't immediately update the capacity model?

None, it's just that the lend token exchange rate is only updated lazily

gregdhill (approved):

PR #866, opened 2023-01-19
fix: make rewards rpc work

sander2:
Without this, the rpc would fail with `Unable to estimate the current reward` because withdrawing the rewards would fail

### DIFF
--- a/standalone/runtime/tests/test_fee_pool.rs
+++ b/standalone/runtime/tests/test_fee_pool.rs
@@ -122,6 +122,47 @@ fn get_vault_collateral(vault_id: &VaultId) -> Amount<Runtime> {
         .unwrap()
 }
 
+fn estimate_vault_reward_rate(vault_id: VaultId) -> Result<UnsignedFixedPoint, DispatchError> {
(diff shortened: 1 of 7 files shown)


standalone/runtime/tests/test_fee_pool.rs
```diff
@@ -122,6 +122,47 @@ fn get_vault_collateral(vault_id: &VaultId) -> Amount<Runtime> {
         .unwrap()
 }
 
+fn estimate_vault_reward_rate(vault_id: VaultId) -> Result<UnsignedFixedPoint, DispatchError> {
```
daniel-savu, line 125: Have you added this function to the integration test to check that the logic in the RPCs works? Guess we really need to refactor the common logic, starting to look scary to maintain

daniel-savu (approved):

standalone/runtime/tests/test_fee_pool.rs
sander2: Yea I will refactor in a follow-up pr

PR #874, opened 2023-01-20
refactor: move config impls to runtime-common

sander2:
Opening this for discussion. I know it's not ideal - it's kind of ugly how the macros access items that are declared at the place of invocation. But I think it's a good step towards sharing more code across the runtimes. Follow-up step would be to move the consts and the shared `parameter_types!` types to runtime-common. 

Possibly in the future we can clean it up a bit by giving the macros an explict argument that contains the input types


sander2:
Closing this due to received review on discord

PR #895, opened 2023-01-26
refactor: use Amounts consistently in loans pallet

sander2:
~~This is WIP, there are still many functions that I'd want to convert to taking/returning `Amount`s, but I wanted to get feedback before I spend time doing that - I don't want to spend the time if you don't want to merge this with the audit going on. I think that once done, though, it'll result in~~ the code being more readable, more concise, and less error prone. It'll be clear what each balance indicates - there will be no more functions that take an underlying_id but the balance of a lend token, and there will be explicit failures when you try to e.g. add lending token amounts the underlying amounts. Plus code can use functions like `underlying.convert_to(lend_token_id)`, instead of querying the exchange rate and calling the correct conversion function

### DIFF
--- a/crates/loans/src/interest.rs
+++ b/crates/loans/src/interest.rs
@@ -74,7 +74,7 @@ impl<T: Config> Pallet<T> {
         let mut total_reserves = Self::total_reserves(asset_id);
         let mut borrow_index = Self::borrow_index(asset_id);
 
-        let util = Self::calc_utilization_ratio(total_cash, total_borrows, total_reserves)?;
+        let util = Self::calc_utilization_ratio(&total_cash, &total_borrows, &total_reserves)?;
         let borrow_rate = market
             .rate_model
             .get_borrow_rate(util)
@@ -87,31 +87,26 @@ impl<T: Config> Pallet<T> {
             let delta_time = now
                 .checked_sub(last_accrued_interest_time)
                 .ok_or(ArithmeticError::Underflow)?;
-            let interest_accumulated =
-                Self::accrued_interest(borrow_rate, total_borrows, delta_time).ok_or(ArithmeticError::Overflow)?;
-            total_borrows = interest_accumulated
-                .checked_add(total_borrows)
-                .ok_or(ArithmeticError::Overflow)?;
-            total_reserves = market
-                .reserve_factor
-                .mul_floor(interest_accumulated)
-                .checked_add(total_reserves)
-                .ok_or(ArithmeticError::Overflow)?;
+            let interest_accumulated = Self::accrued_interest(borrow_rate, &total_borrows, delta_time)?;
+            total_borrows = interest_accumulated.checked_add(&total_borrows)?;
+            total_reserves = interest_accumulated
+                .map(|x| market.reserve_factor.mul_floor(x))
+                .checked_add(&total_reserves)?;
 
             borrow_index = Self::increment_index(borrow_rate, borrow_index, delta_time)
                 .and_then(|r| r.checked_add(&borrow_index))
                 .ok_or(ArithmeticError::Overflow)?;
         }
 
-        let exchange_rate = Self::calculate_exchange_rate(total_supply, total_cash, total_borrows, total_reserves)?;
+        let exchange_rate = Self::calculate_exchange_rate(&total_supply, &total_cash, &total_borrows, &total_reserves)?;
 
         Ok((
             borrow_rate,
             supply_rate,
             exchange_rate,
             util,
-            total_borrows,
-            total_reserves,
+            total_borrows.amount(),
+            total_reserves.amount(),
             borrow_index,
         ))
     }
@@ -125,27 +120,24 @@ impl<T: Config> Pallet<T> {
         let total_borrows = Self::total_borrows(asset_id);
         let total_reserves = Self::total_reserves(asset_id);
 
-        Self::calculate_exchange_rate(total_supply, total_cash, total_borrows, total_reserves)
+        Self::calculate_exchange_rate(&total_supply, &total_cash, &total_borrows, &total_reserves)
     }
 
     /// Calculate the borrowing utilization ratio of the specified market
     ///
     /// utilizationRatio = totalBorrows / (totalCash + totalBorrows − totalReserves)
     pub(crate) fn calc_utilization_ratio(
-        cash: BalanceOf<T>,
-        borrows: BalanceOf<T>,
-        reserves: BalanceOf<T>,
+        cash: &Amount<T>,
+        borrows: &Amount<T>,
+        reserves: &Amount<T>,
     ) -> Result<Ratio, DispatchError> {
         // utilization ratio is 0 when there are no borrows
         if borrows.is_zero() {
             return Ok(Ratio::zero());
         }
-        let total = cash
-            .checked_add(borrows)
-            .and_then(|r| r.checked_sub(reserves))
-            .ok_or(ArithmeticError::Overflow)?;
+        let total = cash.checked_add(&borrows)?.checked_sub(&reserves)?;
 
-        Ok(Ratio::from_rational(borrows, total))
+        Ok(Ratio::from_rational(borrows.amount(), total.amount()))
     }
 
     /// The exchange rate should be greater than the `MinExchangeRate` storage value and less than
@@ -169,11 +161,19 @@ impl<T: Config> Pallet<T> {
         })
     }
 
-    fn accrued_interest(borrow_rate: Rate, amount: BalanceOf<T>, delta_time: Timestamp) -> Option<BalanceOf<T>> {
-        borrow_rate
-            .checked_mul_int(amount)?
-            .checked_mul(delta_time.into())?
+    fn accrued_interest(
+        borrow_rate: Rate,
+        amount: &Amount<T>,
+        delta_time: Timestamp,
+    ) -> Result<Amount<T>, DispatchError> {
+        let balance = borrow_rate
+            .checked_mul_int(amount.amount())
+            .ok_or(ArithmeticError::Overflow)?
+            .checked_mul(delta_time.into())
+            .ok_or(ArithmeticError::Overflow)?
             .checked_div(SECONDS_PER_YEAR.into())
+            .ok_or(ArithmeticError::Underflow)?;
+        Ok(Amount::new(balance, amount.currency()))
     }
 
     fn increment_index(borrow_rate: Rate, index: Rate, delta_time: Timestamp) -> Option<Rate> {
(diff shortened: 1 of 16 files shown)


crates/loans/src/lib.rs
```diff
@@ -1295,9 +1295,11 @@ pub mod pallet {
             let from = Self::incentive_reward_account_id();
             Self::ensure_active_market(asset_id)?;
             Self::accrue_interest(asset_id)?;
-            let exchange_rate = Self::exchange_rate_stored(asset_id)?;
-            let voucher_amount = Self::calc_collateral_amount(redeem_amount, exchange_rate)?;
-            let redeem_amount = Self::do_redeem_voucher(&from, asset_id, voucher_amount)?;
+
+            let redeem_amount = Amount::new(redeem_amount, asset_id);
+            let voucher = redeem_amount.convert_to(Self::lend_token_id(asset_id)?)?;
```
daniel-savu, line 1300: This is much cleaner, though one thing to mention is that `CurrencyConvert` tries to re-accrue interest. If interest has already been accrued in this block it's a no-op but I wasn't sure if there would be any edge cases caused by always re-accruing.

crates/loans/src/lib.rs
```diff
@@ -1359,24 +1361,16 @@ impl<T: Config> Pallet<T> {
...
-        lend_token_amount: BalanceOf<T>,
+        voucher: &Amount<T>,
     ) -> Result<Amount<T>, DispatchError> {
-        let effects_amount = Self::collateral_balance(asset_id, lend_token_amount)?;
-        Self::get_asset_value(asset_id, effects_amount)
+        let asset_id = Self::underlying_id(voucher.currency())?;
+
+        let underlying = voucher.convert_to(asset_id)?;
+        let market = Self::market(asset_id)?;
+        let effects = underlying.map(|x| market.collateral_factor.mul_ceil(x));
```
daniel-savu, line 1371: LGTM overall, this does touch lots of the critical logic and would've been good to get this in before the audit. The multiplication on this line is not hard to miss, for example

sander2: Yea the audit is my main point of concern. We could ask them if they mind changing commit versions - it hasn't been long since kick-off so it's possible they haven't looked into it too much. Alternatively, we could ask them to review the pr afterwards.

gregdhill (commented):
Seems to improve readability in some places, others we anyway convert back to `.amount()` which is slightly less clean. Happy to go with this approach if you think it makes sense @sander2.

sander2:
> Seems to improve readability in some places, others we anyway convert back to .amount() which is slightly less clean. 

That was still WIP. Now `.amount()` should pretty much only be used when
- emitting events,
- writing to storage
- doing the explicit currency conversion

It would be cleaner if we also had `Amount`s in storage/events but I didn't want this to be a breaking change

crates/loans/src/tests/interest_rate.rs
```diff
@@ -7,20 +7,28 @@ use sp_runtime::{
     traits::{CheckedDiv, One, Saturating},
     FixedPointNumber,
 };
+use traits::OracleApi;
 
 #[test]
 fn utilization_rate_works() {
+    let f = |x| Amount::new(x, Token(KSM));
```
daniel-savu, line 14:
```suggestion
    let ksm = |x| Amount::new(x, Token(KSM));
```

crates/loans/src/types.rs
```diff
@@ -59,7 +59,7 @@ impl<T: Config> AccountLiquidity<T> {
 #[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug, Default, TypeInfo)]
 pub struct BorrowSnapshot<Balance> {
     /// Principal Total balance (with accrued interest), after applying the most recent balance-changing action
-    pub principal: Balance,
+    pub principal: Balance, // amount of underlying borrowed, to be paid back eventually
```
daniel-savu, line 62: Can move this comment to the doc comment on the line above

crates/loans/src/lend_token.rs
```diff
@@ -141,7 +141,10 @@ impl<T: Config> Pallet<T> {
         let reducible_underlying_amount = reducible_supply_amount.convert_to(underlying_id)?.amount();
 
         let exchange_rate = Self::exchange_rate(underlying_id);
-        let amount = Self::calc_collateral_amount(reducible_underlying_amount, exchange_rate)?;
+        let amount = FixedU128::from_inner(reducible_underlying_amount)
```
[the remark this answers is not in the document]
sander2: Because it doesn't exist anymore - I had removed it since no code seemed to depend on it anymore after my refactoring. But then it turned out this test code still used it.. I didn't want to re-add the function since no non-testing code should use it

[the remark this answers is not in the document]
daniel-savu: You could also move it to `lend_token.rs` where the other test utilities are

gregdhill, line 144: How come you don't need to convert `reducible_underlying_amount` using `checked_from_int`?

crates/loans/src/lib.rs
```diff
@@ -482,7 +502,6 @@ pub mod pallet {
     /// If an account minted without also locking their balance as collateral, their corresponding entry
     /// in this map will be zero.
     #[pallet::storage]
-    #[pallet::getter(fn account_deposits)]
```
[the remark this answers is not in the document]
gregdhill: Looks like @sander2 added new methods here: https://github.com/interlay/interbtc/pull/895/files#diff-3361e3a76ed6ed3f01d210d7f33acac1a1616e6a93a2b82fd043cd8607735932R1337-R1351

crates/loans/src/lib.rs
```diff
@@ -1348,47 +1384,32 @@ impl<T: Config> Pallet<T> {
...
-            if currency_borrow_amount.is_zero() {
+            let borrow = Self::current_borrow_balance(borrower, asset_id)?;
+            if borrow.is_zero() {
                 continue;
             }
-            total_borrow_value =
-                Self::get_asset_value(asset_id, currency_borrow_amount)?.checked_add(&total_borrow_value)?;
+            let value = Self::get_asset_value(&borrow)?;
+            total_borrow_value.checked_accrue(&value)?;
         }
```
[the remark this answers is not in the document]
gregdhill: It's still readable, don't think too much into it :stuck_out_tongue:

crates/loans/src/lib.rs
```diff
@@ -2169,16 +2125,21 @@ impl<T: Config> LoansTrait<CurrencyId<T>, AccountIdOf<T>, BalanceOf<T>, Amount<T
...
+    // NOTE: used in OracleApi, so don't use oracle calls here or it'll recurse forever
     fn recompute_collateral_amount(underlying: &Amount<T>) -> Result<Amount<T>, DispatchError> {
         // This function could be called externally to this pallet, with interest
         // possibly not having accrued for a few blocks. This would result in using an
         // outdated exchange rate. Call `accrue_interest` to avoid this.
         Self::ensure_active_market(underlying.currency())?;
         Self::accrue_interest(underlying.currency())?;
         let exchange_rate = Self::exchange_rate_stored(underlying.currency())?;
-        let underlying_amount = Self::calc_collateral_amount(underlying.amount(), exchange_rate)?;
+        let lend_amount = FixedU128::from_inner(underlying.amount())
```
gregdhill, line 2136: Minor nit: `lend_amount` seems less clear to me than `underlying_amount`

gregdhill: Also same concern with `FixedU128` conversion (or lack thereof) probably not an issue but just want to make sure

crates/loans/src/lib.rs
```diff
@@ -2109,47 +2066,43 @@ impl<T: Config> LoansTrait<CurrencyId<T>, AccountIdOf<T>, BalanceOf<T>, Amount<T
...
-    ) -> Result<(), DispatchError> {
+    fn do_redeem(supplier: &AccountIdOf<T>, amount: &Amount<T>) -> Result<(), DispatchError> {
+        let asset_id = amount.currency();
         Self::ensure_active_market(asset_id)?;
         Self::accrue_interest(asset_id)?;
-        let exchange_rate = Self::exchange_rate_stored(asset_id)?;
-        let voucher_amount = Self::calc_collateral_amount(amount, exchange_rate)?;
-        let redeem_amount = Self::do_redeem_voucher(supplier, asset_id, voucher_amount)?;
+
+        let voucher = amount.to_lend_token()?;
```
gregdhill, line 2094: Nice refactoring :smile:

crates/loans/src/lib.rs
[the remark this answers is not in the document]
sander2: Indeed, I added getters that return `Amount`s

crates/loans/src/mock.rs
```diff
@@ -232,7 +228,10 @@ pub fn with_price(
             (_, currency) if currency == DEFAULT_WRAPPED_CURRENCY => {
                 return MockResult::Return(Ok(Amount::new(amount.amount(), DEFAULT_WRAPPED_CURRENCY)));
             }
-            (currency, _) if currency == DEFAULT_WRAPPED_CURRENCY => {
+            (currency, x) if currency == DEFAULT_WRAPPED_CURRENCY => {
+                return MockResult::Return(Ok(Amount::new(amount.amount(), x)));
+            }
+            (a, b) if a == b => {
```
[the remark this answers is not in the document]
sander2: It's needed because this identity conversion is common here: https://github.com/interlay/interbtc/blob/1a1afa90228f37c9ade4acbda8275c2f5cfe85ce/crates/currency/src/lib.rs#L67

E.g. `Amount::new(1, KSM).convert_to(LEND_KSM)` hits this case

crates/loans/src/types.rs
sander2: I really did not understand the line above fwiw, but yea I'll delete

crates/loans/src/lend_token.rs
sander2: Because it's `from_inner` rather than `from`. So a value of `1` results in a value of 1e-18

crates/loans/src/lib.rs
sander2: > Minor nit: `lend_amount` seems less clear to me than `underlying_amount`

Yea except that it's not the underlying amount :) it's returning lend tokens, not underlying tokens. It was misnamed previously

crates/loans/src/lib.rs
sander2: Why would there by edge cases if it's a no-op?

crates/loans/src/interest.rs
```diff
@@ -169,11 +161,19 @@ impl<T: Config> Pallet<T> {
...
-        borrow_rate
-            .checked_mul_int(amount)?
-            .checked_mul(delta_time.into())?
+    fn accrued_interest(
+        borrow_rate: Rate,
+        amount: &Amount<T>,
+        delta_time: Timestamp,
+    ) -> Result<Amount<T>, DispatchError> {
+        let balance = borrow_rate
+            .checked_mul_int(amount.amount())
```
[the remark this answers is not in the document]
sander2: Can you create an issue? I'd like to keep this pr free from functional changes if possible

crates/loans/src/lib.rs
```diff
@@ -1732,43 +1727,36 @@ impl<T: Config> Pallet<T> {
...
 
-        let incentive_reserved = market.liquidate_incentive_reserved_factor.mul_floor(
-            FixedU128::from_inner(collateral_amount)
-                .checked_div(&market.liquidate_incentive)
-                .map(|r| r.into_inner())
-                .ok_or(ArithmeticError::Underflow)?,
-        );
-        let incentive_reserved_amount: Amount<T> = Amount::new(incentive_reserved, lend_token_id);
+        let incentive_reserved = amount_to_liquidate
+            .checked_div(&market.liquidate_incentive)?
```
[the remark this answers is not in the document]
sander2: It's not really meant to be a bugfix haha. In the old code, the `collateral_amount` and `amount_to_liquidate` have the same value: https://github.com/interlay/interbtc/blob/1a1afa90228f37c9ade4acbda8275c2f5cfe85ce/crates/loans/src/lib.rs#L1715-L1717

This change is just meant to be a refactoring to make use of the `checked_div` inside the `Amount`

crates/loans/src/lib.rs
daniel-savu: The side effect would be accruing interest more often, which is good for the lending protocol but will result in slightly higher fees

daniel-savu (approved):

gregdhill (approved):

PR #897, opened 2023-01-27
Single-block loans via batching

daniel-savu:
There is currently zero interest that is applied to a loan that is paid back in the same block, because interest is only accrued in the first lending transaction of a block. This has other implications and probably needs to change.

This lag in accruing also allows for taking interest-free loans if they are batched after a very small one (until someone else triggers an interest rate accrual in the market).


daniel-savu:
Closing, this is acceptable behaviour because over-collateralization is still required to do this so overall this adds liquidity to the market

PR #899, opened 2023-01-28
feat: adding to vault-id to events #898

jtfirek:
My solution to obtain the information I need on the squid side of things #898.
This is my first time writing/reading production rust so I might be off here. I would love to get on a call and chat with someone about it!


jtfirek:
I have found another event I can listen to that has the information I need.

PR #900, opened 2023-01-30
test(loans): batched loans accrue interest

daniel-savu:
Ensures that batched loans accrue full interest, even though interest is lazily accrued.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5144,6 +5144,7 @@ dependencies = [
  "orml-tokens",
  "orml-traits",
  "pallet-timestamp",
+ "pallet-utility",
  "parity-scale-codec",
  "scale-info",
  "serde",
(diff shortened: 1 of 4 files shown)


sander2 (approved):
Fyi unless you do something like `run_to_block`, there isn't really any different afaik between batching 2 calls and making 2 calls sequentially, since no blocks are actually being produced. But doesn't hurt to do it this way either and it makes the intent clear

daniel-savu:
> Fyi unless you do something like `run_to_block`, there isn't really any different afaik between batching 2 calls and making 2 calls sequentially, since no blocks are actually being produced. But doesn't hurt to do it this way either and it makes the intent clear

I assumed so, just wanted to come up with a realistic PoC of a hack if the second loan were to not accrue interest

Issue #902, opened 2023-01-31
Round up `checked_mul_int` in `accrued_interest`


daniel-savu:
**Describe the bug**
Even though the error caused by rounding individual interest rates is [now](https://github.com/interlay/interbtc/pull/896) made in favour of the protocol, the lend token exchange rate does not benefit from this. To fix, replace `checked_mul_int` with an approach that rounds up in the `accrued_interest` function.

Issue #901, opened 2023-01-30
Switch from simple interest to compound interest calculation in lending


daniel-savu:
Compound's [audit by OpenZeppelin](https://blog.openzeppelin.com/compound-audit/) reported that interest is accrued using the simple interest formula, as opposed to the compound interest formula, which lowers the interest accrued long-term if there aren't frequent interactions with a market.

Switch to using compound interest in the `accrued_interest` and `increment_index` functions of the `loans` pallet.

PR #904, opened 2023-02-01
fix(loans): Use compound interest formula

daniel-savu:
Found a better fix that initially planned for https://github.com/interlay/interbtc/issues/902, which also closes https://github.com/interlay/interbtc/issues/901.

- `accrued_interest` can be removed altogether by reusing the interest accrual function for total borrows. Using `saturating_sub` instead of `checked_sub` [here](https://github.com/interlay/interbtc/blob/ba7b2344503afdff6c003ebc03737309772d1990/crates/loans/src/lib.rs#L1533) is actually correct, because if `total_borrows` is always rounded up then it will exceed the sum of all debt owed by borrowers. This is undesirable because it would cause the internal exchange rate to be more generous to lenders than it should, because lend tokens could be redeemed for more cash than the protocol has (or will have after debt repayment). By rounding down, the exchange rate increases more slowly than it should while loans are active, but it catches up with its "true value" as repayments are made because more cash becomes available in the protocol than `total_borrows` approximated.
    - It revealed a bigger problem though, caused by using the simple interest formula instead of the compound interest one. Since the borrow snapshot only gets updated when a user borrows or repays, but the total borrow index gets updated on each market interaction (possibly every block), the latter gets compounded while the former does not, reducing the debt a borrower actually needs to repay. 
- replaces usage of simple interest formula by the compound interest formula, to fix the above. Note that this causes several hardcoded values in the tests to increase because interest is now correctly compounded. Without this, the less frequent a market is interacted with, the less interest borrowers owe.

Closes https://github.com/interlay/interbtc/issues/901
Closes https://github.com/interlay/interbtc/issues/902

### DIFF
--- a/crates/loans/src/interest.rs
+++ b/crates/loans/src/interest.rs
@@ -88,9 +88,7 @@ impl<T: Config> Pallet<T> {
             let delta_time = now
                 .checked_sub(last_accrued_interest_time)
                 .ok_or(ArithmeticError::Underflow)?;
-            borrow_index_new = Self::increment_index(borrow_rate, borrow_index, delta_time)
-                .and_then(|r| r.checked_add(&borrow_index))
-                .ok_or(ArithmeticError::Overflow)?;
+            borrow_index_new = Self::accrue_index(borrow_rate, borrow_index, delta_time)?;
             let total_borrows_old = total_borrows.clone();
             total_borrows =
                 Self::borrow_balance_from_old_and_new_index(&borrow_index, &borrow_index_new, total_borrows)?;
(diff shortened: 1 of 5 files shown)


crates/loans/src/interest.rs
```diff
@@ -161,26 +171,17 @@ impl<T: Config> Pallet<T> {
...
+    pub(crate) fn accrue_index(borrow_rate: Rate, index: Rate, delta_time: Timestamp) -> Result<Rate, DispatchError> {
+        // Compound interest:
+        // new_index = old_index * (1 + annual_borrow_rate / SECONDS_PER_YEAR) ^ delta_time
+        let rate_fraction = borrow_rate
             .checked_div(&FixedU128::saturating_from_integer(SECONDS_PER_YEAR))
+            .ok_or(ArithmeticError::Underflow)?;
+        let base_rate = Rate::one()
+            .checked_add(&rate_fraction)
+            .ok_or(ArithmeticError::Overflow)?;
+        let compounded_rate = base_rate.saturating_pow(usize::saturated_from(delta_time));
```
sander2, line 183: What's the unit of delta_time? Asking because pow will panic if it's > 2^32

philippb90 (approved):
The interest rate computation looks good to me from what I understand.

crates/loans/src/lib.rs
```diff
@@ -1551,16 +1548,35 @@ impl<T: Config> Pallet<T> {
...
+    pub fn borrow_balance_from_old_and_new_index(
+        old_index: &FixedU128,
+        new_index: &FixedU128,
+        amount: Amount<T>,
+        rounding_mode: RoundingMode,
+    ) -> Result<Amount<T>, DispatchError> {
+        // Calculate new borrow balance using the interest index:
+        // recent_borrow_balance = snapshot.principal * borrow_index / snapshot.borrow_index
+        let borrow_index_increase = new_index.checked_div(&old_index).ok_or(ArithmeticError::Underflow)?;
+        let borrow_balance = match rounding_mode {
```
sander2, line 1570: will refactor this after #908

crates/loans/src/lib.rs
```diff
@@ -1551,16 +1548,35 @@ impl<T: Config> Pallet<T> {
...
+            RoundingMode::Up => amount.checked_fixed_point_mul_rounded_up(&borrow_index_increase)?,
+            RoundingMode::Down => {
+                let balance_u128 = borrow_index_increase
+                    .checked_mul_int(amount.amount())
+                    .ok_or(ArithmeticError::Overflow)?;
+                Amount::<T>::new(balance_u128, amount.currency())
+            }
+        };
```
sander2, line 1578: will refactor this after #908

sander2 (approved):
As discussed on discord, we're merging now to start testnet testing with this. More thorough review of this pr, including performance implications are TODO

PR #923, opened 2023-02-09
chore(loans): Update weights

daniel-savu:
Since interest is now accrued using `saturating_pow`, whose complexity is `O(log(delta_seconds))`, this PR recomputes weights. However the `delta_seconds` are passed as a `usize` which is 32 bits in WASM, so `saturating_pow` can be considered constant time. Most weights have decreased with the exception of `withdraw_all_collateral`, which now re-accrues interest in the currency conversion function. `liquidate_borrow` and `reduce_incentive_reserves` are still hardcoded, pending the addition of the oracle pallet to the mock runtime.


sander2:
> so `saturating_pow` can be considered constant time

Technically true, but I don't think the benchmark is hitting the worst case. But we can address that in another pr

PR #930, opened 2023-02-14
fix: release correct amount of underlying in loans.redeem

### LINKED ISSUE #929: Loans: redeem not releasing correct amount due to rounding
In both `redeem`, the underlying amount is first converted to the lend token amount, after which the to-be released collateral is calculated by converting the lend token back to the underlying amount. This can have the consequence that the released collateral is less than what was requested in the `redeem` call. We should refactor to avoid the second conversion

sander2:
This fixes an issue where I called `loans.redeem` with `1e8` as amount, but only `1e8 - 1` actually got released. This was due to the amount first being converted to the lend_token_id and then back to the underlying. The change is that we now only do one conversion: in `loans.redeem` we do only a single `to_lend_id`, while in `loans.redeem_all` we do only a single `to_underlying_id`.  

Closes #929

### DIFF
--- a/crates/currency/src/amount.rs
+++ b/crates/currency/src/amount.rs
@@ -224,6 +224,11 @@ mod math {
             Ok(if self.le(other)? { self.clone() } else { other.clone() })
         }
 
+        pub fn max(&self, other: &Self) -> Result<Self, DispatchError> {
+            ensure!(self.currency_id == other.currency_id, Error::<T>::InvalidCurrency);
+            Ok(if self.ge(other)? { self.clone() } else { other.clone() })
+        }
+
         pub fn lt(&self, other: &Self) -> Result<bool, DispatchError> {
             ensure!(self.currency_id == other.currency_id, Error::<T>::InvalidCurrency);
             Ok(self.amount < other.amount)
--- a/crates/loans/src/lib.rs
+++ b/crates/loans/src/lib.rs
@@ -1012,14 +1012,17 @@ pub mod pallet {
             let who = ensure_signed(origin)?;
             ensure!(!redeem_amount.is_zero(), Error::<T>::InvalidAmount);
 
-            let amount = Amount::<T>::new(redeem_amount, asset_id);
+            let underlying = Amount::<T>::new(redeem_amount, asset_id);
+            let voucher = underlying.to_lend_token()?;
 
-            Self::do_redeem(&who, &amount, &amount.to_lend_token()?)?;
+            Self::do_redeem(&who, &underlying, &voucher)?;
 
             Ok(().into())
         }
 
         /// The caller redeems their entire lend token balance in exchange for the underlying asset.
+        /// Note: this will fail if the account needs some of the collateral for backing open borrows,
+        /// or if any of the lend tokens are used by other pallets (e.g. used as vault collateral)
         ///
         /// - `asset_id`: the asset to be redeemed.
         #[pallet::call_index(11)]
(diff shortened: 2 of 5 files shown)


crates/loans/src/lib.rs
```diff
@@ -1033,42 +1029,14 @@ pub mod pallet {
...
-            // TODO: investigate whether it is possible to implement the conversion functions
-            // with guarantees that this always holds:
-            // `calc_underlying_amount(calc_collateral_amount(x)) = calc_collateral_amount(calc_underlying_amount(x))`
-            // Use the `converting_to_and_from_collateral_should_not_change_results` unit test to achieve this.
-            // If there are leftover lend_tokens after a `redeem_all` (because of rounding down), it would make it
-            // impossible to enforce "collateral toggle" state transitions.
-            Self::ensure_active_market(asset_id)?;
+            let free = Self::free_lend_tokens(asset_id, &who)?;
+            let reserved = Self::reserved_lend_tokens(asset_id, &who)?;
+            let used_collateral = Amount::max(&free, &reserved)?;
```
daniel-savu, line 1034: An account can't have both `free` and `reserved` lend tokens so this is correct but is not very easy to understand at first glance. Would be good to add a comment.

daniel-savu: @sander2 Actually the caller could have both `free` and `reserved` lend tokens if they just mint and then lock some of their balance as vault collateral. If they lock more vault collateral than the free collateral, this call will end up failing although it should just redeem the free tokens.

sander2: > @sander2 Actually the caller could have both `free` and `reserved` lend tokens if they just mint and then lock some of their balance as vault collateral. If they lock more vault collateral than the free collateral, this call will end up failing although it should just redeem the free tokens.

Oh shit I was assuming that free and reserved would be mutually exclusive.. That's not hard to fix but that means I reviewed some of the code with incorrect assumptions :(

crates/loans/src/lib.rs
```diff
@@ -1007,68 +1012,31 @@ pub mod pallet {
...
-            // but that would cause the `accrue_interest_works_after_redeem_all` unit test to fail with
-            // left: `1000000000003607`,
-            // right: `1000000000003608`'
-
-            // Chaining `calc_underlying_amount` and `calc_collateral_amount` continuously decreases
-            // an amount because of rounding down, and having the current function call `do_redeem`
-            // would perform three conversions: lend_token -> token -> lend_token -> token.
-            // Calling `do_redeem_voucher` directly only performs one conversion: lend_token -> token,
-            // avoiding this edge case.
-            // TODO: investigate whether it is possible to implement the conversion functions
```
daniel-savu, line 1046: I would leave this TODO somewhere in the code

crates/currency/src/amount.rs
```diff
@@ -224,6 +224,11 @@ mod math {
             Ok(if self.le(other)? { self.clone() } else { other.clone() })
         }
 
+        pub fn max(&self, other: &Self) -> Result<Self, DispatchError> {
```
daniel-savu, line 227: This function can be removed now

sander2: I chose to leave it since it's a potentially useful utility function

daniel-savu (approved):

crates/loans/src/lib.rs
sander2: Created https://github.com/interlay/interbtc/issues/934

PR #943, opened 2023-02-22
[BREAKING (metadata-only)] refactor: loans cleanup

sander2:
This will affect metadata, meaning clients may get subxt errors


daniel-savu:
I was considering this but I'm not sure whether we can just set the market state to "Pending" instead, when we want to halt the market. Good to check with a test.

sander2:
> I was considering this but I'm not sure whether we can just set the market state to "Pending" instead, when we want to halt the market. Good to check with a test.

I didn't think about the possibility of wanting the halt the market, but to be fair halting is not possible with the current code anyway, right? Barring a raw storage write

sander2:
Not really a prio right now, better not to change things needlessly rn

PR #947, opened 2023-02-22
refactor: remove reserves accounting

sander2:
The pallet compiles, but I did not update tests yet because I wanted feedback _on the general idea_ before I spend time doing that. What this pr does, is to remove all references to the `Reserves`. Instead, we just transfer accumulated interest directly to the treasury, thus removing the need to keep explicit bookkeeping.

### DIFF
--- a/crates/loans/src/interest.rs
+++ b/crates/loans/src/interest.rs
@@ -35,12 +35,17 @@ impl<T: Config> Pallet<T> {
             return Ok(());
         }
 
-        let (borrow_rate, supply_rate, exchange_rate, util, total_borrows_new, total_reserves_new, borrow_index_new) =
+        let (borrow_rate, supply_rate, exchange_rate, util, total_borrows_new, reserve_interest, borrow_index_new) =
             Self::get_market_status(asset_id)?;
 
         Self::update_last_accrued_interest_time(asset_id, now)?;
+
+        reserve_interest
+            .transfer(&Self::account_id(), &T::Treasury::get())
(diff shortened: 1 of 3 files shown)


daniel-savu:
The reserve is not just a place to store treasury-owned funds, but also a way of providing liquidity to the market (when properly implemented). It's true that we could just have proposals to deposit treasury funds into lending via the regular way, but the reserve approach would keep the liquidity in by default and treasury would withdraw only when really needed.

daniel-savu:
The other consideration is that reserve liquidity does not earn the treasury interest, so it could be argued that it is against the protocol's own interest do use the reserve feature. Depositing via minting qTokens does earn the treasury interest.

sander2:
> The reserve is not just a place to store treasury-owned funds, but also a way of providing liquidity to the market (when properly implemented). It's true that we could just have proposals to deposit treasury funds into lending via the regular way, but the reserve approach would keep the liquidity in by default and treasury would withdraw only when really needed.

> The other consideration is that reserve liquidity does not earn the treasury interest, so it could be argued that it is against the protocol's own interest do use the reserve feature. Depositing via minting qTokens does earn the treasury interest.

Hmm I see what you mean. The current implementation doesn't make the rewards available for other people borrow though. I guess it's a design choice whether or not treasury rewards should be put as liquidity by default. Did you talk with Dom or Alexei about this? Another downside is that treasury may not be able to withdraw their rewards at all times if people happen to be loaning it.

PR #970, opened 2023-03-14
test: liquidation with liquidate_incentive_reserved_factor=0

sander2:
The only usage of `liquidate_incentive_reserved_factor` I could find was in `liquidated_transfer`, which only gets called from `liquidate_borrow` as far as I can tell. This test checks that having it set to 0 doesn't cause any problems

### DIFF
--- a/crates/loans/src/tests/liquidate_borrow.rs
+++ b/crates/loans/src/tests/liquidate_borrow.rs
@@ -1,12 +1,12 @@
 use crate::{
     mock::{
         new_test_ext, with_price, CurrencyConvert, Loans, RuntimeOrigin, Test, Tokens, _run_to_block, market_mock,
-        new_test_ext_no_markets, ALICE, BOB, DEFAULT_WRAPPED_CURRENCY, LEND_KBTC, LEND_KSM,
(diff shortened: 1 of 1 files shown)


crates/loans/src/tests/liquidate_borrow.rs
```diff
@@ -154,6 +154,72 @@ fn full_workflow_works_as_expected() {
...
+        assert_eq!(Tokens::balance(KSM, &ALICE), unit(1100),);
+        assert_eq!(Loans::account_borrows(KSM, ALICE).principal, unit(50));
+        assert_eq!(Tokens::balance(KSM, &BOB), unit(750));
+        assert_eq!(
+            Loans::exchange_rate(KBTC).saturating_mul_int(Tokens::balance(Loans::lend_token_id(KBTC).unwrap(), &BOB)),
+            unit(110),
+        );
+        // 0 dollar reserved in our incentive reward account
+        let incentive_reward_account = Loans::incentive_reward_account_id();
+        println!("incentive reserve account:{:?}", incentive_reward_account.clone());
```
daniel-savu, line 211: remove?

sander2: That was copied from the existing test but sure

daniel-savu (approved):

PR #971, opened 2023-03-17
feat(loans): add mutation testing decorations

daniel-savu:
Decorates all functions in the loans pallet with `#[cfg_attr(test, mutate)]`, so `mutagen` can mutate them and check if the unit tests break, "killing" the mutations. Only unit tests are used when checking, so some of the surviving mutants exist because they're tested in an integration test.

Most surviving mutants just remove the `deposit_event` call, however some logic in the interest rate model should actually be better tested. This is the output about surviving mutants: [SURVIVED.txt](https://github.com/interlay/interbtc/files/11003404/SURVIVED.txt)

The `mutagen` dependency uses the latest master on their GitHub, because version `2.x` wasn't released to crates.io for some reason.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4511,6 +4511,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "json"
+version = "0.12.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


sander2:
Related: a long time ago we also looked into getting code coverage info in the CI. I remember it worked, but we never integrated it

PR #981, opened 2023-03-24
fix: limit number of swaps in dexGeneral

sander2:
Quick fix so we can close https://github.com/interlay/srlabs-audit/issues/10. Longer term we should switch the boundedvec, but that will be a breaking change


gregdhill (dismissed):

daniel-savu (approved):
LGTM, should apply this limiting in `dex-stable` as well, if I understand correctly

PR #995, opened 2023-04-05
Farming benches v2

daniel-savu:


crates/farming/src/benchmarking.rs
```diff
@@ -62,62 +62,88 @@ pub fn get_benchmarking_currency_ids() -> Vec<(CurrencyId, CurrencyId)> {
...
 
-benchmarks! {
-    on_initialize {
-        let c in 0 .. get_benchmarking_currency_ids().len() as u32;
+#[benchmarks]
+pub mod benchmarks {
+    use super::*;
+
+    #[benchmark]
+    pub fn on_initialize(c: Linear<1, 4>) {
```
daniel-savu, line 70: I hardcoded the upper bound - it previously was `get_benchmarking_currency_ids().len()` but now needs to be a constant. Maybe we could generate the benchmarking currency array using `ForeignAsset`?

gregdhill (approved):

PR #1021, opened 2023-04-20
Loans toggle migration

daniel-savu:
- Migrates inconsistent lend-token storage (both `free` and `reserved`) if any is found, by manually calling `lock_if_account_deposited(...)` (what the missing hooks were supposed to do)
- Sets orml-token hooks for `loans` in the Kintsugi runtime ~~(I assume this only fixes for a new genesis though - how do I set the config of a live runtime?)~~
- Bumps Kintsugi spec from `1023002` to `1023003`
- Adds TODO for setting the orml-tokens hooks before the interlay launch. Maybe easier to track as a GitHub issue in the Interlay DeFi Hub milestone?
- Tested using `try-runtime` on Kintsugi mainnet (`RUST_LOG=info target/release/interbtc-parachain try-runtime --wasm-execution=compiled --chain=kintsugi-latest --runtime target/release/wbuild/kintsugi-runtime-parachain/kintsugi_runtime_parachain.wasm on-runtime-upgrade live --uri 'wss://api-kusama.interlay.io:443/parachain'`)
- One issue I'm having is that even though the `Debug` trait is implemented by the `AccountId` and `CurrencyId` types, the logs show empty values: `Locked all free collateral for , in <wasm:stripped>` (should be `Locked all free collateral for [account_id], in [currency_id]`).

### DIFF
--- a/crates/loans/src/lib.rs
+++ b/crates/loans/src/lib.rs
@@ -34,7 +34,7 @@ use frame_support::{
     log,
     pallet_prelude::*,
     require_transactional,
-    traits::{tokens::fungibles::Inspect, UnixTime},
+    traits::{tokens::fungibles::Inspect, OnRuntimeUpgrade, UnixTime},
     transactional, PalletId,
 };
 use frame_system::pallet_prelude::*;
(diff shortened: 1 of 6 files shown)


crates/loans/src/lib.rs
```diff
@@ -437,33 +438,14 @@ pub mod pallet {
...
-            let max_exchange_rate = crate::MaxExchangeRate::<T>::get();
-            if max_exchange_rate.is_zero() {
-                crate::MaxExchangeRate::<T>::put(Rate::from_inner(DEFAULT_MAX_EXCHANGE_RATE));
-            }
-            let min_exchange_rate = crate::MinExchangeRate::<T>::get();
-            if min_exchange_rate.is_zero() {
-                crate::MinExchangeRate::<T>::put(Rate::from_inner(DEFAULT_MIN_EXCHANGE_RATE));
-            }
-            T::DbWeight::get().reads_writes(2, 2)
+            migration::collateral_toggle::Migration::<T>::on_runtime_upgrade()
```
sander2, line 441: Since you also added the migration to `Executive`, the migration will run twice. Also, it'll run for each runtime. I don't think that's a problem in your particular migration but with it already being in `Executive`, it shouldn't be required here

sander2 (commented):
> Sets orml-token hooks for loans in the Kintsugi runtime (I assume this only fixes for a new genesis though - how do I set the config of a live runtime?)

No you assume incorrectly :) The hooks are part of the runtime, which will be upgraded when we do the runtime upgrade. The only things that are only set at genesis are things configured by the chain spec (and the initialization of newly added pallets, I think)

crates/loans/src/lib.rs
sander2: Let's remove here, I'll double check with chopsticks that the migration runs

crates/loans/src/migration.rs
```diff
@@ -0,0 +1,148 @@
...
+                inconsistent_balances.len() == 0,
+                "There should be no inconsistent lending collateral balances left"
+            );
+            Ok(())
+        }
+    }
+}
+
+#[cfg(test)]
+#[cfg(feature = "try-runtime")]
```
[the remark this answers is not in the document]
daniel-savu: I did run the test, yes. Also ran `try-runtime` on the kintsugi mainnet and it fixes the issue. What I didn't check is whether the hooks work once the new runtime is in operation.

crates/loans/src/lib.rs
```diff
@@ -437,10 +437,6 @@ pub mod pallet {
 
     #[pallet::hooks]
     impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
-        fn on_runtime_upgrade() -> frame_support::weights::Weight {
```
sander2, line 439: I'd have removed the whole hooks implementation, rn it still has pre and post hooks. But whatever, I don't think it needs to block merging

sander2 (approved):

PR #1023, opened 2023-04-21
chore: orml-tokens benchmarks

sander2:
Note that I couldn't use the benchmark in interlay runtime since it doesn't contain lending


gregdhill (approved):
Would be nice if we could somehow parameterize the weights for `orml-tokens` based on the `CurrencyId` so that only certain operations are more expensive.

sander2:
> Would be nice if we could somehow parameterize the weights for `orml-tokens` based on the `CurrencyId` so that only certain operations are more expensive.

We certainly could do that but since fees are pretty low, I see this is low prio - I'd say let's keep as-is until there is an actual demand to decrease fees

PR #1048, opened 2023-05-04
chore: benchmarks&weights for extrinsic&block overhead

sander2:
See:  https://github.com/paritytech/substrate/blob/master/utils/frame/benchmarking-cli/src/overhead/README.md

To run:
```
target/release/interbtc-parachain benchmark overhead --chain interlay-testnet-bench --execution=wasm --wasm-execution=compiled --warmup 100 --repeat 10 --weight-path parachain/runtime/testnet-interlay/src/weights/
target/release/interbtc-parachain benchmark overhead --chain kintsugi-testnet-bench --execution=wasm --wasm-execution=compiled --warmup 100 --repeat 10 --weight-path parachain/runtime/testnet-kintsugi/src/weights/
target/release/interbtc-parachain benchmark overhead --chain kintsugi-bench --execution=wasm --wasm-execution=compiled --warmup 100 --repeat 10 --weight-path parachain/runtime/kintsugi/src/weights/
target/release/interbtc-parachain benchmark overhead --chain interlay-bench --execution=wasm --wasm-execution=compiled --warmup 100 --repeat 10 --weight-path parachain/runtime/interlay/src/weights/
```


gregdhill (dismissed):
Assuming the templates for these files are given to us from Substrate? Will we also need to run the `overhead` command on production hardware? Can't imagine we need to test this at the PR level but we should remember to add this to whatever release procedure we adopt.

sander2:
See https://github.com/interlay/interbtc/pull/1048/commits/710984ead5861e9723bf7598d5edac9f2057ec64 , I allowed a small fee increase.. which is not ideal, but this also happened every time we upgrade polkadot versions

gregdhill (approved):

PR #1054, opened 2023-05-12
feat: allow payment in non-native currencies by swapping with dex-general

sander2:
# Existing solutions

I had a look at how other projects implement bring-your-own-fee. Specifically, I looked at Acala, HydraDX and substrate.

# Existing implementations

## Substrate native: pallet-asset-tx-payment

Substrate contains the `pallet-asset-tx-payment` pallet. This pallet changes the extrinsic header to include a CurrencyId. When a transaction is processed, a price source is used to convert the fee amount to the given currency. However, it does not do any swaps - it just pays the block author the equivalent amount in the given currency. This pallet is used by e.g. statemine and composable. The support in polkadot.js explorer is not good. Signing will work but it doesn't allow you to select different curriences. I _think_ we could select the currency in our our front-end code, though.

## HydraDX

Hydra uses a custom pallet. Like the substrate one, it does not do any swaps, and just pays the block author in a different currency. The documentation states otherwise, but it is incorrect - I checked on-chain behavior: see e.g. block 2363593. Unlike the substrate pallet, they configure the fee currency not on a per-extrinsic level, but instead they configure the used currency in a dedicated extrinsic (and as a consequence, it's persistent). Implementation detail: they use the same `pallet_transaction_payment` that we do, but with `OnChargeTransaction` that they implement in their `transaction-multi-payment` pallet. 

## Acala

Acala uses their own pallet, but unlike the previous two solutions, Acala actually performs a swap into their native currency. Similarly to Hydra, they have extrinsics to set the used currency (and swap path), but they also have a single-shot `with_fee_path` function. Their implementation is overkill for our purposes - they have this thing called a fee pool (in addition to the dex) and they allow fees to be paid by other accounts. They also reimplement a lot of code provided by substrate. 

# Our implementation

One of the requirements given is to use actual swaps, which means that Acala's solution is closest to our requirements. However, for our purposes it is needlessly complex, so what I chose to do was to use their approach of `with_fee_path`, but with our own implementation. I did my best to keep our implementation as simple as possible, since this feature requires code execution prior to dispatch, which if you're not careful can easily lead to bugs that allow DoS attacks. It took a couple of iterations but I'm pretty happy with the simplicity of the current result. It reuses as much of the substrate implementation as possible.

The usage of this pallet is as follows. If a user want to pay fees in a different currency, rather than calling `tx`, they call `multiTransactionPayment.withFeeSwapPath(path, ..., tx)`. UI will need to wrap all calls like this - the swap path is not committed to storage for future transactions (see discord discussion).

## How it works

First, a short summary of how transactions are processed. In our runtimes, we provide the SignedExtras, which is a tuple of `SignedExtension`s. In this case, the significant on is `pallet-transaction-payment`: https://github.com/paritytech/substrate/blob/fd4c7b38c952cbac826e2da0a61ea40965f4296a/frame/transaction-payment/src/lib.rs#L780-L843 . Extensions have three functions: `validate`, `pre_dispatch` and `post_dispatch`. When a node receives a transaction, it first calls `validate`. If this function returns an error, the transaction does not get included in the transaction queue, so it will never get included in a block. Typically the `validate` and `pre_dispatch` have very similar implementation; the first has to check if the second will be succesful. It also assigns a priority based on the tip left by the user. Anyway, if a transaction passes the all the `validate` checks, and it gets included into a block, the `pre_dispatch` gets executed, in which pallet-transaction-payment calls `OnChargeTransaction::withdraw_fee`. After the dispatch, `post_dispatch` gets called, which calls `OnChargeTransaction::correct_and_deposit_fee` to refund unused fees (i.e. if weight is less than expected, or if the function returns `Pays::No`).

I implemented `OnChargeTransaction` on our newly added pallet. In the `withdraw_fee`, which is called from `validate` and `pre_dispatch`, we inspect the call, and if the call is `with_fee_swap_path`, it will do the swap. Since the swap is done before the actual dispatch, `with_fee_swap_path` itself is not doing any swaps. The only thing it does, is to dispatch the wrapped call, and to return the correct weight returned by it.

The weight of `with_fee_swap_path` is set to the weight of the wrapped call, plus the weight of the swap. This ignores the logic for calculating the fee, and for checking the swap path. I think it's probably insignificant, but we could benchmark that part if needed.

Note that if the fees are lower than expected due to the wrapped call returning a weight override, or it returning `Pays::No`, then the surplus is returned in the native currency; no additional swap gets done to swap it back to the fee currency.

For now, I only added it to the kintsugi runtime. If this passes review I can add to the others as well.

Chopsticks transaction from an account that only has KSM:
image


crates/multi-transaction-payment/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    #[pallet::pallet]
+    pub struct Pallet<T>(_);
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight({
+			let dispatch_info = call.get_dispatch_info();
+			((T::DexWeightInfo::swap_assets_for_exact_assets(_path.len() as u32)).saturating_add(dispatch_info.weight), dispatch_info.class,)
```
sander2, line 70: Slightly inaccurate since it neglects the weight of executing `modify_weight`, but it's only a tiny bit of code so I think it's fine for now. I'll add an issue to address this in the future

crates/multi-transaction-payment/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+            let mut ret = call.dispatch(origin);
+            // modify any returned weight overrides. Note that we pass-through the `Pays` unmodified.
+            let modify_weight = |x: &mut PostDispatchInfo| {
+                x.actual_weight = x
+                    .actual_weight
+                    .map(|x| x.saturating_add(T::DexWeightInfo::swap_assets_for_exact_assets(_path.len() as u32)));
+            };
+            match ret {
+                Ok(ref mut info) => {
+                    // if a weight override is returned, add the cost of the swap
```
daniel-savu, line 91: If this code compiled it means that the return value of the dispatch is always a `PostDispatchInfo`, right? In that case I'd remove the `if` in this comments to avoid confusion
```suggestion
                    // a weight override is returned, add the cost of the swap
```

crates/multi-transaction-payment/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+        #[pallet::call_index(0)]
+        #[pallet::weight({
+			let dispatch_info = call.get_dispatch_info();
+			((T::DexWeightInfo::swap_assets_for_exact_assets(_path.len() as u32)).saturating_add(dispatch_info.weight), dispatch_info.class,)
+		})]
+        #[frame_support::transactional]
+        pub fn with_fee_swap_path(
+            origin: OriginFor<T>,
+            _path: Vec<T::CurrencyId>,
+            _amount_in_max: <T as currency::Config>::Balance,
```
daniel-savu, line 76: Why is this called `_amount_in_max`? It looks like the entire amount is converted, even though the unspent fee is returned in the native currency

daniel-savu: Oh, I see that this signature matches `inner_swap_assets_for_exact_assets` and I assume `fee` is calculated in `pre_dispatch`

crates/multi-transaction-payment/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+    impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight({
+			let dispatch_info = call.get_dispatch_info();
+			((T::DexWeightInfo::swap_assets_for_exact_assets(_path.len() as u32)).saturating_add(dispatch_info.weight), dispatch_info.class,)
+		})]
+        #[frame_support::transactional]
+        pub fn with_fee_swap_path(
+            origin: OriginFor<T>,
+            _path: Vec<T::CurrencyId>,
```
daniel-savu, line 75: Any reason not to use a `BoundedVec` here? Otherwise the logic for checking the swap path could be a source of a DoS attack

crates/multi-transaction-payment/src/mock.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+}
+
+pub struct CurrencyConvert;
+impl currency::CurrencyConversion<currency::Amount<Test>, CurrencyId> for CurrencyConvert {
+    fn convert(
+        _amount: &currency::Amount<Test>,
+        _to: CurrencyId,
+    ) -> Result<currency::Amount<Test>, sp_runtime::DispatchError> {
+        unimplemented!()
```
daniel-savu, line 133: Paying with lend tokens would increase the max weight, but not by much

crates/multi-transaction-payment/src/tests.rs
```diff
@@ -0,0 +1,417 @@
...
+            let result = call.dispatch(RuntimeOrigin::signed(who));
+            let post_info = match result {
+                Ok(post_info) => post_info,
+                Err(err) => err.post_info,
+            };
+
+            Payment::post_dispatch(Some(pre), &info, &post_info, len, &Ok(())).unwrap();
+
+            BalanceChange {
+                native_cost: (initial_native_balance as i128) - (Tokens::free_balance(native_currency, &who) as i128),
```
daniel-savu, line 114: I would expect this value to always be less than or equal to zero, if a correct amount of `fee_balance` is converted into the native currency

crates/multi-transaction-payment/src/tests.rs
```diff
@@ -0,0 +1,417 @@
...
+        let who = 1;
+        let native_currency = GetNativeCurrencyId::get();
+
+        let call = RuntimeCall::Testing(testing_helpers::Call::weighted {
+            expected: Weight::from_ref_time(expected_weight),
+            actual_weight: actual_weight.map(|x| Weight::from_ref_time(x)),
+            err,
+            pays_fee,
+        });
+        dry_run_given_call(who, native_currency, native_currency, pays_fee, call, 0).native_cost as u128
```
daniel-savu, line 130: This should always be less than or equal to zero, should the return type should be signed

crates/multi-transaction-payment/src/tests.rs
```diff
@@ -0,0 +1,417 @@
...
+}
+
+#[test]
+/// basic test that payment in native fee continues to work as normal
+fn pay_in_native_fee_succeeds() {
+    new_test_ext().execute_with(|| {
+        // <Payment as SignedExtension>::pre_dispatch();
+        let payment_extension = pallet_transaction_payment::ChargeTransactionPayment::<Test>::from(1);
+        let who = 1;
+        let call = RuntimeCall::System(frame_system::Call::remark { remark: vec![] });
```
daniel-savu, line 215: Similar to this, is it worth testing a call wrapped in `with_fee_swap_path` where the path has one or two (duplicate) entries of the native currency?

crates/multi-transaction-payment/src/tests.rs
```diff
@@ -0,0 +1,417 @@
...
+        assert_ok!(Tokens::set_balance(
+            RuntimeOrigin::root(),
+            who,
+            fee_currency,
+            initial,
+            0
+        ));
+
+        // check that validate executes successfully, but don't commit storage
+        dry_run(|| {
```
daniel-savu, line 365: Could this test not be refactored to use `dry_run_given_call` if that takes an additional closure to check assertions?

crates/multi-transaction-payment/src/tests.rs
```diff
@@ -0,0 +1,417 @@
...
+/// test that tips are paid in the fee currency, not the native currency
+fn pay_in_swapped_fee_with_tip() {
+    new_test_ext().execute_with(|| {
+        let baseline = dry_run_swapped_with_tip(6, None, false, Pays::Yes, 0);
+        assert!(baseline.fee_cost > 0); // decrease in fee currency, no change in native
+        assert_eq!(baseline.native_cost, 0);
+
+        // Paying tip increases the amount of fee taken in the fee currency
+        let cost = dry_run_swapped_with_tip(6, None, false, Pays::Yes, 1);
+        assert!(cost.fee_cost > baseline.fee_cost);
```
daniel-savu, line 318: Is there a way to know the expected amount resulting from a conversion, including slippage? These inequality assertions seem rather weak

crates/multi-transaction-payment/src/lib.rs
sander2: The return value is always `PostDispatchInfo` indeed, but `PostDispatchInfo.actual_weight` is an Option - it may or may not return a weight override

crates/multi-transaction-payment/src/lib.rs
sander2: It's called `_amount_in_max` because it's passed along to `inner_swap_assets_for_exact_assets` as argument. It allows you to set an upper limit to the amount of the input tokens to prevent cases where you'd pay a lot more than expected due to being frontrun by someone in the dex. `fee` (which is in the native currency) is determined by the `pallet_transaction_payment` crate

crates/multi-transaction-payment/src/mock.rs
sander2: That should be part of the dex swap weight

crates/multi-transaction-payment/src/lib.rs
sander2: It's a pass-through to the dex-general swapping function, which is using a Vec rather than a boundedVec. It shouldn't be a DOS vulnerability since the weight of the dex swapping function is parametrized by the vector length

crates/multi-transaction-payment/src/tests.rs
sander2: This function makes no assumption on the fee currency - it can be the native token as well (i.e. without swapping), meaning it can be positive or negative. That's why it's i128 rather than u128

crates/multi-transaction-payment/src/tests.rs
sander2: This is dry-running a call using the native currency as fee, i.e. not swap is being performed. So the cost will be 0 or positive

crates/multi-transaction-payment/src/tests.rs
sander2: I don't think so - all this pallet cares about is that the path ends in the native token. If the dex can't handle paths with duplicate entries, it should have unit tests for that inside the dex pallet itself

crates/multi-transaction-payment/src/tests.rs
sander2: `dry_run_given_call` takes a `RuntimeCall`, i.e. an extrinsic, which `validate` isn't. And anyway I think the current solution is quite nice :)

crates/multi-transaction-payment/src/lib.rs
daniel-savu: Then maybe move this comment inside the `modify_weight` closure?

crates/multi-transaction-payment/src/tests.rs
daniel-savu: What I mean is that this entire test (`pay_fee_in_other_currency_succeeds`) follows almost exactly the same pattern as `dry_run_given_call`, except for checking correctness with `assert!(balance < initial);`. No big deal though

daniel-savu (approved):

nud3l (approved):

PR #1061, opened 2023-05-19
feat: issue griefing collateral in selectable currency

sander2:
To accompany https://github.com/interlay/interbtc/pull/1054 

I tested the migration with chopsticks.

### DIFF
--- a/crates/issue/src/lib.rs
+++ b/crates/issue/src/lib.rs
@@ -17,6 +17,8 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
+pub mod migration;
+
 #[cfg(test)]
 extern crate mocktopus;
 
@@ -77,6 +79,7 @@ pub mod pallet {
             amount: BalanceOf<T>,
             fee: BalanceOf<T>,
             griefing_collateral: BalanceOf<T>,
+            griefing_currency: CurrencyId<T>,
             vault_id: DefaultVaultId<T>,
             vault_address: BtcAddress,
             vault_public_key: BtcPublicKey,
@@ -201,9 +204,10 @@ pub mod pallet {
             origin: OriginFor<T>,
             #[pallet::compact] amount: BalanceOf<T>,
             vault_id: DefaultVaultId<T>,
+            griefing_currency: CurrencyId<T>,
         ) -> DispatchResultWithPostInfo {
             let requester = ensure_signed(origin)?;
-            Self::_request_issue(requester, amount, vault_id)?;
+            Self::_request_issue(requester, amount, vault_id, griefing_currency)?;
             Ok(().into())
         }
 
@@ -296,6 +300,7 @@ impl<T: Config> Pallet<T> {
         requester: T::AccountId,
         amount_requested: BalanceOf<T>,
         vault_id: DefaultVaultId<T>,
+        griefing_currency: CurrencyId<T>,
     ) -> Result<H256, DispatchError> {
         let amount_requested = Amount::new(amount_requested, vault_id.wrapped_currency());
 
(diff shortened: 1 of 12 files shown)


crates/issue/src/lib.rs
```diff
@@ -316,11 +321,8 @@ impl<T: Config> Pallet<T> {
         ext::vault_registry::ensure_not_banned::<T>(&vault_id)?;
 
         // calculate griefing collateral based on the total amount of tokens to be issued
-        let amount_collateral = amount_requested.convert_to(T::GetGriefingCollateralCurrencyId::get())?;
```
daniel-savu, line 319: Shouldn't you also remove the `GetGriefingCollateralCurrencyId` associated type from the vault-registry? It's supposed to not be used at all any longer, and that will help point out if other parts of the code should also be changed

daniel-savu: I noticed that `nomination`, `redeem` and `replace` still use it and it's no longer present in the `issue` pallet, so I guess it's fine. Would be cleaner to have the same griefing collateral everywhere though, but that's a product decision.

crates/issue/src/migration.rs
```diff
@@ -0,0 +1,209 @@
...
+use sp_core::H256;
+
+/// The log target.
+const TARGET: &'static str = "runtime::issue::migration::v1";
+
+/// The original data layout of the democracy pallet without a specific version number.
+mod v0 {
+    use super::*;
+
+    // Due to a known bug in serde we need to specify how u128 is (de)serialized.
```
daniel-savu, line 12: Does this comment still apply? I can't find this deserialization configured anywhere

crates/issue/src/migration.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    pub struct Migration<T>(sp_std::marker::PhantomData<T>);
+
+    impl<T: Config + frame_system::Config<Hash = H256>> OnRuntimeUpgrade for Migration<T> {
+        #[cfg(feature = "try-runtime")]
+        fn pre_upgrade() -> Result<Vec<u8>, &'static str> {
+            assert_eq!(StorageVersion::get::<Pallet<T>>(), 0, "can only upgrade from version 0");
+
+            let issue_count = v0::IssueRequests::<T>::get().len();
+            log::info!(target: TARGET, "{} public proposals will be migrated.", issue_count,);
```
daniel-savu, line 60: nit
```suggestion
            log::info!(target: TARGET, "{} issue requests will be migrated.", issue_count,);
```

crates/issue/src/migration.rs
daniel-savu: Ah it's because the v0 fields are annotated with macros

crates/issue/src/lib.rs
```diff
@@ -348,6 +350,7 @@ impl<T: Config> Pallet<T> {
             amount: amount_user.amount(),
             fee: fee.amount(),
             griefing_collateral: griefing_collateral.amount(),
+            griefing_currency: griefing_collateral.currency(),
```
daniel-savu, line 353: Why not just change the type of `griefing_collateral` from `Balance` to `Amount`, to include `griefing_currency`?

daniel-savu: I'm assuming that it messes up storage byte alignment

crates/issue/src/migration.rs
```diff
@@ -0,0 +1,209 @@
...
+    use super::*;
+    use crate::{
+        tests::{Test as T, *},
+        types::*,
+    };
+    use frame_support::bounded_vec;
+
+    #[allow(deprecated)]
+    #[test]
+    fn migration_works() {
```
daniel-savu, line 128: This does not test anything related to the issue request migration

standalone/runtime/tests/test_issue.rs
```diff
@@ -667,101 +682,110 @@ mod execute_pending_issue_tests {
...
-                    vault.to_be_issued -= requested_btc;
-                })
-            );
-
-            // issue request is updated: status is complete
-            assert_eq!(
-                IssuePallet::issue_requests(issue_id).unwrap().status,
-                IssueRequestStatus::Completed
-            );
+            for griefing_currency in [DEFAULT_GRIEFING_CURRENCY, vault_id.collateral_currency()] {
```
daniel-savu, line 685: According to the implementation, the griefing currency could be set to anything, not just vault collateral currencies. Not saying this test is wrong but it seems to assume that the griefing currency will not be e.g. `IBTC`

daniel-savu (approved):
My comments are not blocking

standalone/runtime/tests/test_issue.rs
sander2: It's main purpose was to test in the default griefing collateral plus one other one, it doesn't really matter which one it is. But maybe it makes more sense to use a hardcoded one

crates/issue/src/lib.rs
sander2: redeem doesn't use it (except for an interaction with to-be-replaced tokens). Replace does use it, and we can change that in the future if we want, but not a prio for now

nud3l (approved):

PR #1126, opened 2023-07-12
feat: add scale encode and decode as traits to CurrencyId.

nakul1010:
The Pr adds the following things, 
- [x] scale encode and decode as a trait to `CurrencyId`, it is a dependency for PR [#483](https://github.com/interlay/interbtc-clients/pull/483), so that `Static` type wrapping can be avoided for `CurrenctId`


sander2 (approved):

sander2:
More details: https://github.com/paritytech/subxt/issues/1040

PR #1201, opened 2023-09-27

nakul1010:
Description Imported from [Pr 589](https://github.com/interlay/interbtc/pull/589/)

2 significant changes: 

- Only award a premium for the amount of tokens required to bring the vault back to the secure threshold. Otherwise the vault could end up paying the premium for _all_ of its tokens.
- Rather than only considering `issued_tokens` to determine whether or not a vault is below the premium redeem threshold, now use the "to_be_backed" tokens, i.e. the tokens if all open issues and redeems execute successfully (= issued + to_be_issued - to_be_redeemed). This is required because otherwise additional redeems would again be able to de a premium redeem. 
- Since the vault will lose collateral by giving out premium redeem fees, consider that collateral as well when calculating premium collateral fees for the user.  

Since this significantly modifies audited code, I'd like to have 2 people approve this PR before we merge
## Example based on test case
- Vault A
    - Thresholds
        - secure: 200%
        - premium: 160%
        - liquidation:  110%
    - secured capacity: 500 Tokens
    - issued_tokens: 650 Tokens
- User redeems 400 tokens
    - Based on first approach: ~~Premium redeem fee should be given on difference_in_tokens_to_reach_secure_threshold +  potential_decrease_in_issued_tokens_after_giving_out_collatral_as_premium_fees = 150 tokens + 3.75 tokens = 153.75 tokens equivalent to 307.5 COL tokens if exchange_rate is 2~~
    - Premium redeem fee should be given on `max_premium_amount` a vault can pay  = 153.84 tokens equivalent to 307.68 COL tokens if exchange_rate is 2
    - Formula for `max_premium_amount`: `SECURE = (oldCol - maxPremium) / ( oldTokens*EXCH - maxPremium/FEE)
`


sander2:
The solution is not 100% correct. But maybe it's good enough.. The problem is what you're doing is:
- Like in #589, determine at how many tokens we'd be at the secure threshold. Determine premium based on that
- Additionally, you now do an additional iteration, where you update the collateral to account for the paid premium, and update the premium based on that.
- The problem is that the step above once again pays out additional premium, thus decreasing the collateral once more. You can do the step above again and again to get ever closer to the correct amount.

The "right" way to do this, is to calculate the maximum amount of premium to payout using math:
``` rust
CollateralizationRate = totalCollateral / convertToCollateral(totalTokens)
// When doing a premium redeem, this changes to:
// NewCollateralization = (oldCol - maxPremium) / ( oldTokens*EXCH - maxPremium/FEE)
// we want NewCollateralization = SECURE threshold, so
SECURE = (oldCol - maxPremium) / ( oldTokens*EXCH - maxPremium/FEE)
// rewriting gives..
maxPremium = (FEE * (oldCol - oldTokens * EXCH * SECURE)) / (FEE -SECURE)
// plugging in number from the test..
maxPremium = (0.05*(2000 - 650 * 2 * 2)) / (0.05 - 2)
= 15.3846153846
```
In comparison, #589 would award max 15, and this PR awards 15.375. Both #589 and this PR are pretty close, so I'm not sure we really need the math based approach. @gregdhill what do you think?

nakul1010:
> The solution is not 100% correct. But maybe it's good enough.. The problem is what you're doing is:
> 
> * Like in [fix: only award redeem premium upto the secure threshold #589](https://github.com/interlay/interbtc/pull/589), determine at how many tokens we'd be at the secure threshold. Determine premium based on that
> * Additionally, you now do an additional iteration, where you update the collateral to account for the paid premium, and update the premium based on that.
> * The problem is that the step above once again pays out additional premium, thus decreasing the collateral once more. You can do the step above again and again to get ever closer to the correct amount.
> 
> The "right" way to do this, is to calculate the maximum amount of premium to payout using math:
> 
> ```rust
> CollateralizationRate = totalCollateral / convertToCollateral(totalTokens)
> // When doing a premium redeem, this changes to:
> // NewCollateralization = (oldCol - maxPremium) / ( oldTokens*EXCH - maxPremium/FEE)
> // we want NewCollateralization = SECURE threshold, so
> SECURE = (oldCol - maxPremium) / ( oldTokens*EXCH - maxPremium/FEE)
> // rewriting gives..
> maxPremium = (FEE * (oldCol - oldTokens * EXCH * SECURE)) / (FEE -SECURE)
> // plugging in number from the test..
> maxPremium = (0.05*(2000 - 650 * 2 * 2)) / (0.05 - 2)
> = 15.3846153846
> ```
> 
> In comparison, #589 would award max 15, and this PR awards 15.375. Both are pretty close, so I'm not sure we really need the math based approach. @gregdhill what do you think?

Changed the approach based on the new formula provided.

crates/redeem/src/lib.rs
```diff
@@ -501,23 +502,48 @@ impl<T: Config> Pallet<T> {
...
         let premium_collateral = if below_premium_redeem {
-            let redeem_amount_wrapped_in_collateral = user_to_be_received_btc.convert_to(currency_id)?;
-            ext::fee::get_premium_redeem_fee::<T>(&redeem_amount_wrapped_in_collateral)?
+            // Calculate the secure vault capacity
+            let secure_vault_capacity = ext::vault_registry::vault_capacity_at_secure_threshold(&vault_id)?;
+            let to_be_backed_tokens = ext::vault_registry::vault_to_be_backed_tokens(&vault_id)?;
+            let difference_in_tokens_to_reach_secure_threshold =
+                to_be_backed_tokens.saturating_sub(&secure_vault_capacity)?;
+
+            if difference_in_tokens_to_reach_secure_threshold.gt(&user_to_be_received_btc)? {
```
sander2, line 515: Why do we need this `if`? Shouldn't the `else` branch always work?

sander2 (commented):
We have a `vaultRegistry_getPremiumRedeemVaults` rpc. @gregdhill I think we should we update that to return only the amount that is premium-redeemable, right?

crates/redeem/src/lib.rs
nakul1010: the formula gives maximum premium, the user shouldn't receive `maximum_premium`, if he has not redeem enough BTC. 
Example, test case: `integration_test_premium_redeem_with_reward_for_full_request`
alice requests 100 BTC redeem, vault can do 150 BTC redeem to reach secure threshold, therefore alice should only receive `premium_collateral` for 100BTC - tx fee ie  98.300 according to testcase, if we remove the if condition he will receive the max `premium_collateral`.

crates/redeem/src/lib.rs
```diff
@@ -501,23 +502,48 @@ impl<T: Config> Pallet<T> {
...
+                let premium_tokens_in_collateral = user_to_be_received_btc.convert_to(currency_id)?;
+
+                ext::fee::get_premium_redeem_fee::<T>(&premium_tokens_in_collateral)?
+            } else {
+                // Formula = max_premium_collateral = (FEE * (oldTokens * EXCH * SECURE - oldCol)) / (SECURE - FEE)
```
sander2, line 520:
```suggestion
                // The goal of premium redeems is to get the vault back the a healthy collateralization ratio. As such, 
                // we only award a premium for the amount of tokens required to get the vault back to secure threshold.
                // The CollateralizationRate is defined as `totalCollateral / convertToCollateral(totalTokens)`
                // When paying a premium, the collateralization rate gets updated according to the following formula:
                //     `NewCollateralization = (oldCol - awardedPremium) / ( oldTokens*EXCH - awardedPremium/FEE)`
                // To calculate the maximum premium we are willing to pay, we set the newCollateralization to
                // the secure threshold, which gives:
                //     `SECURE = (oldCol - awardedPremium) / (oldTokens*EXCH - awardedPremium/FEE)``
                // We can rewrite this formula to calculate the `premium` amount that would get us to the secure threshold:
                //     `maxPremium = (oldTokens * EXCH * SECURE - oldCol) * (FEE / (SECURE - FEE))`
                // Which can be interpreted as:
                //     `maxPremium = missingCollateral * (FEE / (SECURE - FEE))
                // Note that to prevent repeated premium redeems while waiting for execution, we use to_be_backed_tokens
                // for `oldCol`, which takes into account pending issues and redeems
```

Can you add a comment like this please

sander2: Note: in the comment I changed the formula compared to the one I gave earlier in the PR. 

Old: `maxPremium = (FEE * (oldCol - oldTokens * EXCH * SECURE)) / (FEE -SECURE)`
New: `maxPremium = (oldTokens * EXCH * SECURE - oldCol) * (FEE / (SECURE - FEE))`

This will result in positive intermediate results, and requires only a single multiplication between amounts and fixedpoint numbers

crates/redeem/src/lib.rs
```diff
@@ -501,23 +502,48 @@ impl<T: Config> Pallet<T> {
...
+
+                let issued_tokens_in_collateral = to_be_backed_tokens.convert_to(currency_id)?; // oldTokens * EXCH
+
+                let token_exchange_value =
+                    issued_tokens_in_collateral.checked_rounded_mul(&secure_threshold, Rounding::NearestPrefUp)?; // oldTokens * EXCH * SECURE
+
+                ext::fee::get_premium_redeem_fee::<T>(
+                    &token_exchange_value.saturating_sub(&backing_collateral)?, // (oldCol - oldTokens * EXCH * SECURE)
+                )? // FEE * (oldTokens * EXCH * SECURE - oldCol))
+                .checked_div(&ext::fee::apply_premium_redeem_discount::<T>(&secure_threshold)?)? // (SECURE - FEE)
```
sander2, line 534: Can we rewrite to something like this?

```rust
let secure_threshold = ext::vault_registry::get_secure_threshold::<T>(&vault_id)?;
let premium_redeem_rate = ext::fee::premium_redeem_reward_rate(); // NOTE: add this function
let required_collateral = ext::vault_registry::required_collateral(vault_id, to_be_backed_tokens); // NOTE: add this function. This is `(oldTokens * EXCH * SECURE)`
let current_collateral = ext::vault_registry::get_backing_collateral(&vault_id)?;

let missing_collateral = required_collateral.checked_sub(current_collateral)?; // this is `(oldTokens * EXCH * SECURE - oldCol)`

let factor = premium_redeem_rate.checked_div(secure_threshold.checked_sub(premium_redeem_rate)?)?;
let max_premium = missing_collateral.checked_mul(factor);

let actual_premium = max_premium.min(redeem_amount_wrapped_in_collateral.checked_mul(premium_redeem_rate));
```

sander2 (commented):
@gregdhill @nud3l Do we want to award a premium up to the _global_ secure threshold, or to the vault's custom threshold? Personally for my vault I would prefer to only award a premium upto the global secure threshold, otherwise setting a higher threshold seemingly increases the risk of losing money through premium payouts

gregdhill:
Agree with using the second approach @sander2, as for your other comments we should only return the amount that is premium redeemable in the RPC and I also agree we should award premium up to the global secure threshold since that is always lower.

crates/fee/src/lib.rs
```diff
@@ -427,6 +427,19 @@ impl<T: Config> Pallet<T> {
         amount.checked_rounded_mul(&<PremiumRedeemFee<T>>::get(), Rounding::NearestPrefUp)
     }
 
+    /// Apply a premium redeem discount to the given unsigned fixed-point value
+    ///
+    /// # Arguments
+    ///
+    /// * `amount` - amount in collateral (at current exchange rate)
```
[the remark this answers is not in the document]
nakul1010: function discarded.

crates/redeem/src/lib.rs
nakul1010: updated

crates/vault-registry/src/tests.rs
```diff
@@ -1036,40 +1041,43 @@ mod get_vaults_below_premium_collaterlization_tests {
...
             let issue_tokens1: u128 = 50;
             let collateral1 = 50;
             add_vault(id1.clone(), issue_tokens1, collateral1);
 
-            // returned
-            let id2 = vault_id(4);
-            let issue_tokens2: u128 = 50;
-            let collateral2 = 49;
-            add_vault(id2.clone(), issue_tokens2, collateral2);
-
```
sander2, line 1060: why did you modify this test?

crates/vault-registry/src/lib.rs
```diff
@@ -774,6 +774,52 @@ impl<T: Config> Pallet<T> {
...
+        // When paying a premium, the collateralization rate gets updated according to the following formula:
+        //     `NewCollateralization = (oldCol - awardedPremium) / ( oldTokens*EXCH - awardedPremium/FEE)`
+
+        // To calculate the maximum premium we are willing to pay, we set the newCollateralization to
+        // the global secure threshold, which gives:
+        //     `SECURE = (oldCol - awardedPremium) / (oldTokens*EXCH - awardedPremium/FEE)``
+        // We can rewrite this formula to calculate the `premium` amount that would get us to the secure
+        // threshold:     `maxPremium = (oldTokens * EXCH * SECURE - oldCol) * (FEE / (SECURE -
+        // FEE))` Which can be interpreted as:
+        //     `maxPremium = missingCollateral * (FEE / (SECURE - FEE))
```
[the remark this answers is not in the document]
nakul1010: Done by cargo fmt

crates/vault-registry/src/lib.rs
```diff
@@ -1886,6 +1974,19 @@ impl<T: Config> Pallet<T> {
         collateral.convert_to(wrapped_currency)?.checked_div(&threshold)
     }
 
+    pub fn vault_capacity_at_secure_threshold(vault_id: &DefaultVaultId<T>) -> Result<Amount<T>, DispatchError> {
```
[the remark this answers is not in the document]
nakul1010: removed

crates/vault-registry/src/tests.rs
nakul1010: because the `max_premium`  calculation also considers `global_secure_threshold` which was not used by `redeemable_tokens` method 


- so start of the test set the secure collateral threshold to 0.001 so that the `add_vault`, method doesn't pass the `ExceedingVaultLimit`  error
- before calling the `get_premium_redeem_vaults` set thresholds back to default value so that `get_premium_redeem_vaults` works as expected.

nakul1010:
@sander2 
- `RedeemTransactionSize` storage is required to calculate `premium_redeem_vaults` in `vault_registery`, but `redeem` pallet can't be imported in `vault_registery` as it create cyclic dependency. 

1) Migration of storage (move `RedeemTransactionSize` from `redeem` to `vault_registery` pallet )
	- Pros
		- no frontend change required
	- Cons
		- migrating storage just for a method used during RPC call doesn't make sense. 
		- the storage ideally should be stored in redeem pallet as related to redeem request. 

2) Change RPC call (move RPC from `vault_registery` to `redeem` pallet )
	- Pros 
		- no migration required 
	- Cons
		- frontend change required to call a different method

went ahead with 2 approach 

cc: @gregdhill

crates/redeem/src/lib.rs
```diff
@@ -501,23 +502,28 @@ impl<T: Config> Pallet<T> {
...
-
         let below_premium_redeem = ext::vault_registry::is_vault_below_premium_threshold::<T>(&vault_id)?;
         let currency_id = vault_id.collateral_currency();
 
         let premium_collateral = if below_premium_redeem {
             let redeem_amount_wrapped_in_collateral = user_to_be_received_btc.convert_to(currency_id)?;
-            ext::fee::get_premium_redeem_fee::<T>(&redeem_amount_wrapped_in_collateral)?
+            let premium_redeem_rate = ext::fee::premium_redeem_reward_rate::<T>();
+            let premium_for_redeem_amount = redeem_amount_wrapped_in_collateral
+                .checked_rounded_mul(&premium_redeem_rate, Rounding::NearestPrefUp)?;
```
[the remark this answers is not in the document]
sander2: That'd probably be safer yea

crates/vault-registry/src/tests.rs
sander2: > because the `max_premium` calculation also considers `global_secure_threshold` which was not used by `redeemable_tokens` method earlier

This function is testing `get_premium_redeem_vaults` - that was also using `global_secure_threshold` before, right?

If premium threshold is set to 100% then vault1 should not be returned and   vault 2 should, that shouldn't have changed

crates/redeem/src/lib.rs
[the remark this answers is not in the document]
nakul1010: Rounded down

nakul1010:
> I'm ok with moving the rpc to the redeem crate. We could keep the rpc name identical s.t. it doesn't require UI changes. Fwiw there is also an option 3 to add a new getter to the vault_registry config trait, but option 2 works for me as well

changed the RPC method name from `redeem_getPremiumRedeemVaults` to `vaultRegistry_getPremiumRedeemVaults` in the latest commit, this will not affect the frontend code.

nud3l:
> > I'm ok with moving the rpc to the redeem crate. We could keep the rpc name identical s.t. it doesn't require UI changes. Fwiw there is also an option 3 to add a new getter to the vault_registry config trait, but option 2 works for me as well
> 
> changed the RPC method name from `redeem_getPremiumRedeemVaults` to `vaultRegistry_getPremiumRedeemVaults` in the latest commit, this will not affect the frontend code.

I think it will require an update in the lib maybe?

sander2:
> > > I'm ok with moving the rpc to the redeem crate. We could keep the rpc name identical s.t. it doesn't require UI changes. Fwiw there is also an option 3 to add a new getter to the vault_registry config trait, but option 2 works for me as well
> > 
> > 
> > changed the RPC method name from `redeem_getPremiumRedeemVaults` to `vaultRegistry_getPremiumRedeemVaults` in the latest commit, this will not affect the frontend code.
> 
> I think it will require an update in the lib maybe?

Yea it for sure affects frontend/lib. Let's keep the old naming as an alias, see https://docs.rs/jsonrpsee-proc-macros/0.20.3/jsonrpsee_proc_macros/attr.rpc.html#method-attribute

crates/vault-registry/src/tests.rs
nakul1010: reverted back the test case, just have set back secure threshold after `add_vault` method since before `get_premium_redeem_vaults` just used `premium_threshold` now the `max_premium` calculation also uses `global_secure_threshold`.

crates/vault-registry/src/tests.rs
```diff
@@ -1036,9 +1039,13 @@ mod get_vaults_below_premium_collaterlization_tests {
...
 
+            // set back secure threshold
+            let secure = UnsignedFixedPoint::checked_from_rational(200, 100).unwrap(); // 200%
+            VaultRegistry::_set_secure_collateral_threshold(DEFAULT_CURRENCY_PAIR, secure);
+
             assert_eq!(
-                VaultRegistry::get_premium_redeem_vaults(),
-                Ok(vec![(id1, wrapped(issue_tokens1)), (id2, wrapped(issue_tokens2))])
+                VaultRegistry::get_premium_redeem_vaults(400_u32),
+                Ok(vec![(id2, wrapped(452)), (id1, wrapped(451))])
```
sander2, line 1048: If issued_tokens is 50, how come the returned amount is 452??

nakul1010: fixed it and updated the mock to pass the correct value.

sander2:
I pushed a small change. The `is_below_premium_threshold` was confusing to me, since unlike the other checks, it uses `to_be_backed` tokens rather than `issued` tokens. So in the tests `is_below_premium_threshold` was true while `is_below_secure_threshold` was false. Changed the naming to `will_be_below_premium_threshold` to avoid that confusion. I also removed the check altogether from the non-test code since it wasn't necessary. By removing the check, this code gets hit by more of the tests so I'm a little bit more confident it's correct. Also this way we're sure that the premium calculation gets accounted for in the benchmarks.

I also made some changes to the testing code

crates/vault-registry/src/lib.rs
```diff
@@ -1588,22 +1659,56 @@ impl<T: Config> Pallet<T> {
...
+                    Self::calculate_inclusion_fee(vault_id.wrapped_currency(), redeem_transaction_size).ok()?;
+
+                let vault_to_burn_tokens = burn_wrap.checked_add(&inclusion_fee).ok()?;
+
+                let redeem_fee = ext::fee::get_redeem_fee_value::<T>();
+                let amount_wrapped = UnsignedFixedPoint::<T>::one().saturating_sub(redeem_fee);
+
+                let request_redeem_tokens_for_max_premium = vault_to_burn_tokens.checked_div(&amount_wrapped).ok()?;
+
+                if Self::ensure_not_banned(&vault_id).is_ok() && !request_redeem_tokens_for_max_premium.is_zero() {
```
nakul1010, line 1710: `will_be_below_premium_threshold` check method has been removed from here, but this generates a corner case since `vault_to_burn_tokens` will always be a nonzero amount, since `inclusion_fee` is a nonzero amount, therefore the rpc is giving out `max_premium` for vaults which are above the `PremiumRedeemThreshold`. 

To avoid this either in the if case `
- will_be_below_premium_threshold` method needs to be present OR
- check `max_premium_in_collateral` is non zero amount

parachain test case where the corner case can be trigger: `integration_test_try_get_premium_vaults_which_are_sufficiently_collateralized`

nakul1010: added back the condition

sander2 (approved):

PR #1212, opened 2023-12-06
feat: align derivation accross ecosystem

mustermeiszer:
## Description
Currently, there is no common ecosystem wide approach to derive accounts from xcm-origins. This PR adds a converter both used by 
* [Kusama Asset-Hub](https://github.com/polkadot-fellows/runtimes/blob/b8df4b06c78fb2fb2d0f081c76e30569f12363c6/system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs#L91)
* [Centrifuge](https://github.com/centrifuge/centrifuge-chain/blob/93ace412bc269522916cc191caede54733921fa0/runtime/centrifuge/src/xcm.rs#L322)
* [Moonbeam](https://github.com/moonbeam-foundation/moonbeam/blob/3d6eb4e1de334c3259631ac15373e130c48c00af/runtime/moonbeam/src/xcm_config.rs#L111)

The converter was introduced by [Gavin Wood](https://github.com/paritytech/polkadot/pull/7329) in order to create a general purpose conversion. The idea behind this converter is that a user `User1` from `ChainA` has the ability to have a unique account on any other chain in the ecosystem. This so called remote account can only be controlled from `User1` from `ChainA`.

## Motivation and Context
Allows to have remote accounts on Interlay for all purposes a native account could be used. Especially useful for when we do not want to deteriorate UX by switching networks in the UI.
 
## How Has This Been Tested?
In production at Centrifuge and Moonbeam for quite some time.

### DIFF
--- a/parachain/runtime/interlay/src/xcm_config.rs
+++ b/parachain/runtime/interlay/src/xcm_config.rs
@@ -39,6 +39,8 @@ type LocationToAccountId = (
     SiblingParachainConvertsVia<Sibling, AccountId>,
     // Straight up local `AccountId32` origins just alias directly to `AccountId`.
     AccountId32Aliases<ParentNetwork, AccountId>,
+    // Generate remote accounts according to polkadot standards
+    primitives::xcm::HashedDescriptionDescribeFamilyAllTerminal<AccountId>,
 );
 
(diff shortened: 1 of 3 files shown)


mustermeiszer:
FYI: https://github.com/polkadot-fellows/runtimes/pull/67#event-11226181967

PR #1215, opened 2024-01-04

nakul1010:
Backport to `interbtc-v1.25` branch.


nakul1010:
PR open the wrong target branch should have been opened with `nakul/backport` branch. Therefore closing it without merging.

PR #1233, opened 2024-03-27
fix(rate_model): increase MAX jump rate and full rate

spazcoin:
Recently all USDT on Polkadot was borrowed, leading to Moonwell USDT borrow rates of 192% and Interlay $492k of USDT max borrowed, resulting in Interlay USDT borrow rate pegged at 45-50%. The negative thing was that there was no remaining USDT lending liquidity for lenders to withdraw their capital.

Interlay referenda 96 was approved to adjust the USDT/USDC lending params to set the jump rate to 20% and the full rate to 150% with optimal utilization of 80%. After community and team discussion, this solution was agreed to use the higher full rate to push lending back down towards the optimal lending utilization.

https://interlay.subsquare.io/democracy/referenda/96

Unfortunately the referenda failed to execute because the 150% proposed full rate exceeded the hard-coded max value of 50% for the full rate.

https://interlay.subscan.io/block/4905046?tab=event&event=4905046-0

In this commit I recommend:
* leave 10% as max base rate
* increase "jump rate" hardcoded max from 30% to 100%. We were already considering 20-30% for USDT.
* increase "full rate" from 50% to 500%.

I definitely support having limits since people might submit a value with wrong number of zeros accidentally. But let's place the limits higher than any "reasonable" value that we might set in governance. I can see us potentially having a market with jump rate 50% and full rate 300%.

### DIFF
--- a/crates/loans/src/rate_model.rs
+++ b/crates/loans/src/rate_model.rs
@@ -113,8 +113,8 @@ pub struct JumpModel {
 
 impl JumpModel {
     pub const MAX_BASE_RATE: Rate = Rate::from_inner(100_000_000_000_000_000); // 10%
-    pub const MAX_JUMP_RATE: Rate = Rate::from_inner(300_000_000_000_000_000); // 30%
-    pub const MAX_FULL_RATE: Rate = Rate::from_inner(500_000_000_000_000_000); // 50%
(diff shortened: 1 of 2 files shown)


sander2:
Thanks! It shouldn't be a problem, but I'll add some basic tests in a separate pr

sander2 (approved):

PR #1253, opened 2025-10-03
chore: AH Migration backport for Kintsugi

tsenovilla:
This PR aims to protect Kintsugi's cross chain transfers from the upcoming AHM. During the migration, the inference of KSM reserves may lead to lost of funds, so we need to deactivate those transfers during the process.

A bit more context about this issue: https://forum.polkadot.network/t/mandatory-action-guide-for-ahm-broken-native-crosschain-transfers/14634

The proposed solution follows the approach of the ORML team implemented here: https://github.com/open-web3-stack/open-runtime-module-library/pull/1033. 

The affected extrinsics in Kintsugi are:
- `xtokens` -> all extrinsics. `xtokens` use a `ReserveProvider` to determine the reserve, so the patch is needed here.
- `pallet_xcm` ->  This old xcm version doesn't automatically infer the reserves, so theoretically it's not affected. However, the `ReserveProvider` defined on `XcmExecutor` isn't being taken into account by the extrinsics belonging to this pallet for some reason I couldn't figure out, but probably related to the old xcm version. I'm afraid that they keep using Kusama as the reserve after the migration which will result on a conflict with the `xtokens` reserve (AssetHub). This might end up creating unexpected issues, so for the sake of security I've disabled transferring KSM using `pallet_xcm` during and after the migration. If the codebase is updated in the future to newer xcm versions this block might be removed, in the meanwhile, please use `xtokens` to transfer KSM.

Please note that all other reserves aren't affected by this patch and keep working as usual. The migration status should be changed by calling the `xtokens.set_migration_phase` extrinsic. This extrinsic is only callable by the root origin on Kintsugi, let us know if you prefer to use a different origin.

As the Interbtc codebase deps are too far in the past, the time doesn't allow to bump everything and bring that change with the ORML crates, so we'll be using an ORML fork from the exact commit used by the Interbtc codebase. The fork is located in the [R0gue](https://r0gue.io/) GitHub organization, concretely here: https://github.com/r0gue-io/open-runtime-module-library/tree/master.

**IMPORTANT**: While this branch contains the patch for both Interlay and Kintsugi (just for convenience, to allow us compiling the whole workspace), it should only be used on Kintsugi yet. The branch sets off from the `1.25.6` tag, so it's not suitable for an Interlay upgrade. However, backporting the patch should be straightforward.

**Note**: I'm pushing this branch against the master branch. I'm not aware of Interbtc upgrades policies, so let me know if I should push it against another branch

### DIFF
--- a/parachain/runtime/runtime-tests/src/relaychain/kusama_cross_chain_transfer.rs
+++ b/parachain/runtime/runtime-tests/src/relaychain/kusama_cross_chain_transfer.rs
@@ -1,7 +1,8 @@
 use crate::relaychain::kusama_test_net::*;
 use codec::Encode;
-use frame_support::assert_ok;
-use orml_traits::MultiCurrency;
+use frame_support::{assert_noop, assert_ok};
+use orml_traits::{location::Reserve, MultiCurrency};
+use orml_xtokens::{AbsoluteReserveProviderMigrationPhase, MigrationPhase};
 use primitives::{
     CurrencyId::{ForeignAsset, Token},
     CustomMetadata, TokenSymbol,
@@ -11,6 +12,11 @@ use xcm::latest::{prelude::*, Weight};
 use xcm_builder::ParentIsPreset;
 use xcm_emulator::{TestExt, XcmExecutor};
 use xcm_executor::traits::Convert;
+
+fn sibling_sovereign_account() -> AccountId {
+    use sp_runtime::traits::AccountIdConversion;
+    polkadot_parachain::primitives::Sibling::from(SIBLING_PARA_ID).into_account_truncating()
+}
 
 mod fees {
     use super::*;
@@ -271,11 +277,6 @@ fn transfer_to_relay_chain() {
 /// integrated.
 #[test]
 fn transfer_to_sibling_and_back() {
-    fn sibling_sovereign_account() -> AccountId {
-        use sp_runtime::traits::AccountIdConversion;
-        polkadot_parachain::primitives::Sibling::from(SIBLING_PARA_ID).into_account_truncating()
-    }
-
     Sibling::execute_with(|| {
         register_kint_as_foreign_asset();
     });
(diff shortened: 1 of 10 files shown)


sander2:
Changed the base to `kintsugi-1.25`, which corresponds to 1.25.6 for easier reviewing

parachain/runtime/runtime-tests/src/relaychain/kusama_cross_chain_transfer.rs
```diff
@@ -691,3 +692,349 @@ fn register_kint_as_foreign_asset() {
...
+            Box::new(MultiLocation::new(1, X1(Junction::AccountId32 { id: BOB, network: None })).into()),
+            WeightLimit::Unlimited
+        ));
+    });
+
+    KusamaNet::execute_with(|| {
+        let bob_balance = kusama_runtime::Balances::free_balance(&AccountId::from(BOB));
+
+        // A little bit less to pay fees
+        assert!(bob_balance < KSM.one());
```
sander2, line 796: should check that it's >0, right?

parachain/runtime/runtime-tests/src/relaychain/kusama_cross_chain_transfer.rs
```diff
@@ -691,3 +692,349 @@ fn register_kint_as_foreign_asset() {
...
+            Box::new(MultiLocation::new(1, X1(Junction::AccountId32 { id: BOB, network: None })).into()),
+            WeightLimit::Unlimited
+        ));
+    });
+
+    KusamaNet::execute_with(|| {
+        let bob_balance = kusama_runtime::Balances::free_balance(&AccountId::from(BOB));
+
+        // A little bit less to pay fees
+        assert!(bob_balance < 2 * KSM.one());
```
sander2: should check > 1 KSM

parachain/runtime/runtime-tests/src/relaychain/kusama_cross_chain_transfer.rs
```diff
@@ -691,3 +692,349 @@ fn register_kint_as_foreign_asset() {
...
+            Box::new(concrete_fungible(MultiLocation::parent()).into()),
+            0
+        ));
+    });
+
+    KusamaNet::execute_with(|| {
+        let bob_balance = kusama_runtime::Balances::free_balance(&AccountId::from(BOB));
+
+        // A little bit less to pay fees
+        assert!(bob_balance < KSM.one());
```
sander2: and >0

parachain/runtime/runtime-tests/src/relaychain/kusama_cross_chain_transfer.rs
tsenovilla: True, I've modified the test

parachain/runtime/runtime-tests/src/relaychain/kusama_cross_chain_transfer.rs
tsenovilla: I'll be pushing right now the corrected tests

parachain/runtime/runtime-tests/src/relaychain/kusama_cross_chain_transfer.rs
tsenovilla: For this one, it took us a lot of time cause we saw that the received balance was indeed 0. I was checking your repo and didn't see any test for `pallet_xcm` transfers, so I created one agains the 'kintsugi-1.25' branch and observed that the tokens weren't received on KSM even with your current runtime.

So, I guess this is something that should be fixed on your `pallet_xcm` configuration in the future. For now, it's not problematic because `pallet-xcm` reserve transfers will be disabled from the migration onwards, you can keep using `xtokens` for the moment, it's working properly.

I'm just getting rid of this test for this PR as it's useless

PR #1254, opened 2025-10-13
chore: AH Migration backport for Interlay

tsenovilla:
This PR aims to protect Interlay's cross chain transfers from the upcoming AHM. During the migration, the inference of DOT reserves may lead to lost of funds, so we need to deactivate those transfers during the process.

A bit more context about this issue: https://forum.polkadot.network/t/mandatory-action-guide-for-ahm-broken-native-crosschain-transfers/14634

The proposed solution follows the approach of the ORML team implemented here: https://github.com/open-web3-stack/open-runtime-module-library/pull/1033.

The affected extrinsics on Interlay are:

- `xtokens` -> all extrinsics. `xtokens` use a ReserveProvider to determine the reserve, so the patch is needed here.
- `pallet_xcm` -> `pallet_xcm` calls [are filtered out](https://github.com/interlay/interbtc/blob/interbtc-v1.25/parachain/runtime/interlay/src/lib.rs#L200-L208) on Interlay, so they're not affected.

Please note that all other reserves aren't affected by this patch and keep working as usual. The migration status should be changed by calling the `xtokens.set_migration_phase` extrinsic. This extrinsic is only callable by the root origin on Interlay.

As the Interbtc codebase deps are too far in the past, the time doesn't allow to bump everything and bring that change with the ORML crates, so we'll be using an ORML fork from the exact commit used by the Interbtc codebase. The fork is located in the [R0gue](https://r0gue.io/) GitHub organization, concretely here: https://github.com/r0gue-io/open-runtime-module-library/tree/master.

**IMPORTANT**: This branch slightly modifies the Kintsugi runtime, just for convenience, to allow us compiling the whole workspace. However it **MUST** be used only to update Interlay

### DIFF
--- a/.github/actions/cache-npm/action.yml
+++ b/.github/actions/cache-npm/action.yml
@@ -1,0 +1,20 @@
+name: 'Cache NPM'
+description: 'Caching npm dependencies'
+inputs:
+  cache-key:
+    description: 'Key to use when caching npm dependencies'
+    required: true
+    default: 'npm-default'
+runs:
+  using: "composite"
+  steps:
+    - name: Cache NPM dependencies
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.npm
+          node_modules
+        key: ${{ runner.os }}-npm-${{ inputs.cache-key }}-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-npm-${{ inputs.cache-key }}-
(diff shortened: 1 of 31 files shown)


al3mart (approved):
This is now ready for the Interlay team to review.

We are pretty confident with the current state of this PR. As long as there is no feedback from your side, we don't plan on pushing any more changes here.

PR #1265, opened 2026-03-12
build(cargo): enable link-time optimization (LTO) for release profile

### LINKED ISSUE #1245: Enable Link-Time Optimization (LTO)
It's not a problem - just an improvement idea.

I noticed that in the `Cargo.toml` file Link-Time Optimization (LTO) for the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve the application's performance a bit.

I suggest enabling LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project since LTO consumes an additional amount of time to finish the compilation routine. If you think that a regular Release build should not be affected by such a change as well, then I suggest adding an additional `dist` or `release-lto` profile where additionally to regular `release` optimizations LTO will also be added. Such a change simplifies life for maintainers and others interested in the project persons who want to build the most performant version of the application. Using ThinLTO should also help to reduce the build-time overhead with LTO. If we enable it on the Cargo profile level, users, who install the application with `cargo install`, will get the LTO-optimized version "automatically". E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48).

Basically, it can be enabled with the following lines:

```
[profile.release]
lto = true
```

**Describe alternatives you've considered**
Leave things as is.

**Additional context**
N/A

anzzyspeaksgit:
## Description
Fixes #1245

As suggested in the issue, this PR enables Link-Time Optimization (`lto = true`) in the `[profile.release]` section of `Cargo.toml`.

This will help reduce the final binary size and likely yield runtime performance improvements without disrupting development build times (as it only targets the `release` profile).

---
🤖 Generated by anzzyspeaksgit (Autonomous AI OSS Contributor)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ incremental = false
 incremental = false
 
 [profile.release]
+lto = true
 incremental = false
 [patch."https://github.com/paritytech/substrate"]
 binary-merkle-tree = { git = "https://github.com/paritytech//substrate", branch = "polkadot-v1.0.0" }


Dustin4444 (commented):
Fixes #1265

